PR #10, opened 2020-07-06
Fix xpath, improve pepxml.roc_curve()

alephreish:
* Fixes xpath bugs in pepxml.roc_curve() that caused a fatal error [XPathEvalError: Invalid predicate]
* Adds "charge" and "tag" keys to the output to make the function usable in real world.

### DIFF
--- a/pyteomics/pepxml.py
+++ b/pyteomics/pepxml.py
@@ -288,26 +288,27 @@ def roc_curve(source):
     Returns
     -------
     out : list
-        A list of ROC points, sorted by ascending min prob.
+        A list of ROC points.
     """
 
     parser = etree.XMLParser(remove_comments=True, ns_clean=True)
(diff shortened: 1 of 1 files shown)


levitsky:
Thank you! This is a legacy function and to be honest I don't think I've ever used it myself.
One minor question: should we maybe convert charge to an integer?

alephreish:
Hi! I wrote my own parser for pepxml in node.js, but recently decided to switch to pyteomics for everyday use - that's when I discovered also roc_curve().
No, "charge" has to remain string as it is a categorical variable and will be used as such in most of the cases. It goes: "all", "2", "3" and "4".

levitsky:
Ah, OK, thanks.

I'm glad you've found pyteomics to be of use and thank you for the contribution!

PR #11, opened 2020-10-06
A minimal Universal Spectrum Identifier parser and PROXI client

mobiusklein:
This PR implements a minimal Universal Spectrum Identifier parser and PROXI client to consume USIs and return spectra from Peptide Atlas, MassIVE, PRIDe, and jPOST.


levitsky:
Thank you! Looks great.

Looking at the test, though, I wonder if there's an easy way to also test the functions without `numpy`.
Also, not sure how much we should care about Python 2 compatibility at this point, but the `<=` comparison on dictionary keys seems to have a different meaning there. The tests pass as long as the dicts are equal, of course.

mobiusklein:
You're right, on Py2 `dict.keys` isn't `Set`-like. I've fixed that, and removed the dependency on `numpy` from the test too.

levitsky:
Ah yes, this is surely the easiest way. I was probably trying to think of something too complex. Thanks again!

PR #14, opened 2020-10-14
Add pyteomics/version, remove VERSION file

### LINKED ISSUE #13: Add a __version__ method to pyteomics
Typically python packages have a *dunder* method for version: `package.__version__`. 

It helps to check and log the version of the package under use in an analysis. However, I could not find any such `pyteomics.__version__` method to access the version. I see that you have a file `VERSION` in the repo and the `setup.py` file accesses the current version information from here.

Please add a `pyteomics.__version__` method to dynamically provide the version of the package.

If however, I have missed out on something or am mistaken please let me know.

Thank you.

levitsky:
This PR closes #13. It adds a module, `pyteomics/version.py`, which defines a `__version__` string and a function that parses it into a tuple similar to `sys.version_info`, with a simple string representation.


mobiusklein (commented):
Feel free to ignore, but I think `version.version` should probably be a runtime constant. You're unlikely to change that unless you're intentionally manipulating `__version__`, right?

```python
version = _VersionInfo(__version__)
```

levitsky:
Good point, thanks. I would probably name it `version_info` though, so that the user doesn't expect a string. Should I rename `__version__` to `version` then?

mobiusklein:
I think `__version__` is expected to be a string still based upon what I've seen. Renaming `version` to `version_info` is fine though.

levitsky:
I'm not sure if `__version__` is easy enough to find (as it's not tab-completed) and I see no real reason for it to be private, even though it is common. But we can provide both.

PR #17, opened 2020-12-01
Fix warning when overriding use_index

paretje:
The warning currently doesn't work as `use_index` is overridden immediately when file is determined to be seeakable. This moves the warning before `use_index` is set.

### DIFF
--- a/pyteomics/auxiliary/file_helpers.py
+++ b/pyteomics/auxiliary/file_helpers.py
@@ -886,21 +886,18 @@ def _check_use_index(source, use_index, default):
             use_index = bool(use_index)
         if isinstance(source, basestring):
             return use_index if use_index is not None else default
-        seekable = True
         if hasattr(source, 'seekable'):
             if not source.seekable():
+                if use_index:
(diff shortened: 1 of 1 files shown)


levitsky (commented):
Thank you for catching this! Now that I'm looking at the fixed code, though, I see a problem with it, and it makes me think that perhaps the way the code is in master was intentional.
You see, I expected that if `source` has a `seekable()` method, then it probably also has a `mode`. If the mode is binary, but `seekable()` returns `False`, then your version of the code will issue two contradicting warnings and eventually just return `True`. The current version, on the other hand, will only issue the warning about `seekable()` (which is good), but it will return `False` for a binary file object, which is not good.
So a fix is needed, but it's going to be more complicated than the suggested change. We probably need to keep track of the overrides that we do in these sequential checks and fail if there is a conflict. What do you think?
If my thinking is correct, I can try to push a change into your branch, or I can try making my own PR and request your review. What would you prefer?

levitsky:
I made changes to `_check_use_index` and added some more tests for the warnings and errors it raises. Hopefully this addresses the issues you had.

PR #20, opened 2020-12-12
Substitute the five predefined XML entities with their characters in ID strings

### LINKED ISSUE #19: XML indexing fails when an XML entity encoded character appears in the identifier attribute
If a character that must be rendered as an XML entity (e.g. `"` to `&quote;`) appears in the the identifier attribute string, when it is written to the byte stream on disk it will look like this:

`<spectrum index="0" defaultArrayLength="0" id="index=0, IonMobility:&quot;0.80540764&quot;">`

The XML indexing scanner will see that string verbatim and store it as-is in the index.

When `lxml.etree.iterparse` yields `Element` objects for us to work with and check for `id` matches, the XML entities are already converted back into their character representation by the underlying C library, so the XML entity is gone.

The solution is to make the XML indexer convert entities before storing the ID.

mobiusklein:
Closes #19

### DIFF
--- a/pyteomics/xml.py
+++ b/pyteomics/xml.py
@@ -745,6 +745,16 @@ class ByteCountingXMLScanner(_file_obj):
     Inheris from :py:class:`pyteomics.auxiliary._file_obj` to support the object-oriented
     :py:func:`_keep_state` interface.
     """
+    entities = {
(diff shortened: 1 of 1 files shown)


levitsky:
Thank you!

PR #24, opened 2021-01-22
Fixed version parsing

donnyyy777:
Accepts version numbers that do not necessarily contain 3 numbers (x.x.x) - can read versions that contain 1 or 2 numbers (x) or (x.x)

### DIFF
--- a/pyteomics/mztab.py
+++ b/pyteomics/mztab.py
@@ -744,7 +744,7 @@ class MzTab(_MzTabParserBase):
                 self.small_molecule_evidence_table.add(tokens[1:])
 
     def _determine_schema_version(self):
-        version_parsed, variant = re.search(r"(?P<schema_version>\d+(?:.\d+(?:.\d+)?)?)(?:-(?P<schema_variant>[MP]))?", str(self.version)).groups()
+        version_parsed, variant = re.search(r"(?P<schema_version>\d+(?:\.\d+(?:\.\d+)?)?)(?:-(?P<schema_variant>[MP]))?", str(self.version)).groups()
         if variant is None:
             variant = "P"
         self.num_version = [int(v) for v in version_parsed.split(".")]
--- a/pyteomics/mztab.py
+++ b/pyteomics/mztab.py
@@ -744,10 +744,13 @@ class MzTab(_MzTabParserBase):
                 self.small_molecule_evidence_table.add(tokens[1:])
 
     def _determine_schema_version(self):
-        version_parsed, _, variant = str(self.version).partition("-")
-        if variant is None or (variant != "M" and variant != "P"):
+        version_parsed, variant = re.search(r"(?P<schema_version>\d+(?:.\d+(?:.\d+)?)?)(?:-(?P<schema_variant>[MP]))?", str(self.version)).groups()
+        if variant is None:
             variant = "P"
         self.num_version = [int(v) for v in version_parsed.split(".")]
+        # Ensure self.num_version is 3-tuple
+        while len(self.num_version) < 3:
+            self.num_version.append(0)
         self.variant = variant
 
     def keys(self):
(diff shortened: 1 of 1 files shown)


donnyyy777:
Hi,

I was trying to use the MzTab parser on an MzTab file, but found out that it threw errors because of a call to search for a version number in the form of x.x.x when the file I was working with had a version number in the form x.x (2 numbers instead of 3). I edited that part such that it wouldn't throw errors if the version number didn't necessarily contain 3 numbers. I made sure to maintain the check that the variant was either "M" or "P".

pyteomics/mztab.py
```diff
@@ -744,8 +744,8 @@ def _parse(self):
                 self.small_molecule_evidence_table.add(tokens[1:])
 
     def _determine_schema_version(self):
-        version_parsed, variant = re.search(r"(?P<schema_version>\d+.\d+.\d+)(?:-(?P<schema_variant>[MP]))?", self.version).groups()
-        if variant is None:
+        version_parsed, _, variant = str(self.version).partition("-")
+        if variant is None or (variant != "M" and variant != "P"):
             variant = "P"
         self.num_version = [int(v) for v in version_parsed.split(".")]
```
mobiusklein, line 750:
```suggestion
        self.num_version = [int(v) for v in version_parsed.split(".")]
        # Ensure self.num_version is 3-tuple 
        while len(self.num_version) < 3:
              self.num_version.append(0)
```

pyteomics/mztab.py
```diff
@@ -744,8 +744,8 @@ def _parse(self):
                 self.small_molecule_evidence_table.add(tokens[1:])
 
     def _determine_schema_version(self):
-        version_parsed, variant = re.search(r"(?P<schema_version>\d+.\d+.\d+)(?:-(?P<schema_variant>[MP]))?", self.version).groups()
```
mobiusklein, line 747: We can keep the regex for defending against non-integer matches using non-capturing optional groups:
`(?P<schema_version>\d+(?:.\d+(?:.\d+)?)?)(?:-(?P<schema_variant>[MP]))?`

mobiusklein (changes requested):
Could you add a step ensuring that `self.num_version` is always a 3-tuple? This way we can always index it in the event someone wants to reason about features. 

The regex for version parsing does have nice properties still as it ensures that the first components are still numerals. You could get the same effect with a slightly more complex regex, but this isn't strictly necessary.

donnyyy777 (commented):
I made the changes you suggested. I agree that using regex is better in this case to parse the version number since it offers a stricter search. I also added the check to make sure that self.num_version is a 3 tuple.

mobiusklein:
Looks good to me. I don't see any more issues. Was the file you were trying to parse an mzTab 1.0 file?

donnyyy777 (commented):
Yes, mzTab file with version 1.0, that's exactly right

pyteomics/mztab.py
```diff
@@ -744,10 +744,13 @@ def _parse(self):
                 self.small_molecule_evidence_table.add(tokens[1:])
 
     def _determine_schema_version(self):
-        version_parsed, variant = re.search(r"(?P<schema_version>\d+.\d+.\d+)(?:-(?P<schema_variant>[MP]))?", self.version).groups()
+        version_parsed, variant = re.search(r"(?P<schema_version>\d+(?:.\d+(?:.\d+)?)?)(?:-(?P<schema_variant>[MP]))?", str(self.version)).groups()
```
levitsky, line 747: Quick question: are we matching any character `.` instead of a literal `\.` here? Should we fortify the pattern?

```suggestion
        version_parsed, variant = re.search(r"(?P<schema_version>\d+(?:\.\d+(?:\.\d+)?)?)(?:-(?P<schema_variant>[MP]))?", str(self.version)).groups()
```
Doesn't matter all that much, because the splitting occurs later.

mobiusklein: You're correct, we should be matching `\.` not `.`. It was a flaw in my original pattern too. If a really malformed version string were tested it would fail at a different location instead of failing to match the regex.

Issue #31, opened 2021-03-20
Unimod doesn't work with SQLAlchemy 1.4.2


levitsky:
The unimod test fails to start with this error on latest SQLAlchemy 1.4.2:

```
ERROR: setUpClass (__main__.UnimodTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/lev/py/pyteomics/tests/test_unimod.py", line 12, in setUpClass
    cls.handle = unimod.Unimod()
  File "/home/lev/py/pyteomics/pyteomics/mass/unimod.py", line 688, in __init__
    self.session = load(_unimod_xml_download_url)
  File "/home/lev/py/pyteomics/pyteomics/mass/unimod.py", line 655, in load
    for model in Base._decl_class_registry.values():
AttributeError: type object 'Base' has no attribute '_decl_class_registry'
```

mobiusklein:
SQLAlchemy is starting to migrate to a new architecture. I just started seeing breakages on CI for 1.4 on other projects. I'll get started on this at least, it should be easier to fix.

PR #32, opened 2021-03-20
Fix Unimod Compatibility With SQLAlchemy 1.4.X 

mobiusklein:
Addresses #31 

With the migration underway at SQLAlchemy from 1.3 to 1.4, internal details that I shouldn't have relied on are changing as they prepare for a major version change. This now explicitly gathers model types without relying on SQLAlchemy doing it for us indirectly.


levitsky:
Thank you!

PR #37, opened 2021-03-29
Draft ProForma implementation

mobiusklein:
This is a draft implementation for reading and writing the ProForma notation for modified amino acid sequences. I worked to avoid adding dependencies here by making additional controlled vocabularies optional unless you try to parse a string that uses them, and then load them lazily from `psims`.

It still needs more documentation (especially about how to interact with some of its implementation details and which feature annexes it supports) and tests. I can likely inherit several of those from https://github.com/topdownproteomics/sdk/blob/master/tests/TopDownProteomics.Tests/ProForma/ProFormaParserTests.cs.

The ProForma specification is going through review now, but there's already discussion of an update to allow multiple modifications at a single position.


mobiusklein:
@levitsky This should finally be ready for review, with the fundamental functionality all in place.

This adds the `parse_proforma` function to parse a string in ProForma 2.0 format into a list of peptide position tokens and a dictionary of additional modification information (unlocalized, ambiguous or labile modifications, global modification rules, and so on), a `to_proforma` function to take that information and turn it _back_ into a ProForma 2.0 string. It also includes a `ProForma` class which layers on a little more behavior like mass calculation, slicing, and searching for tags by ID.

The non-user-facing bits include all the baroque machinery for dealing with six different modification vocabularies, a more forgiving tokenizer, and a slightly borrowed test suite.

There is still some documentation to iron out, especially which "implementation level" this counts as, as it implements everything but inter-peptide cross-linking support. There's also how to make the users aware of how to control how additional controlled vocabularies are loaded. Right now it uses Unimod directly from `pyteomics.mass.Unimod`, but tries to import `psims` to load the rest, emitting an error message if it needs one of those databases and `psims` isn't installed.

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+        return self.resolver(*keys)
+
+
+class FormulaModification(ModificationBase):
+    prefix_name = "Formula"
+
+    _tag_type = TagTypeEnum.formula
+
+    def resolve(self):
+        # The handling of fixed isotopes is wrong here as Pyteomics uses a different
```
levitsky, line 564: Not only is it wrong, but the formula is currently unsalvageable because the parser swallows all the square brackets.
`'Formula:[13C2][12C-2]H2N'` becomes `'13C212C-2H2N'`, where two numbers `2` and `12` are concatenated and the information is lost. Would there be a way to prevent the parser from consuming the `[` and `]` when they are actually parts of the data (formula) and not markup? Then it shouldn't be too hard to re-format the formula to be correctly parsed by pyteomics into a composition.

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+from pyteomics.mass import Composition, std_aa_mass, Unimod
+from pyteomics.auxiliary import PyteomicsError, BasicComposition
+from pyteomics.auxiliary.utils import add_metaclass
+
+# To eventually be implemented with pyteomics port?
+try:
+    from psims.controlled_vocabulary.controlled_vocabulary import (load_psimod, load_xlmod, load_gno, obo_cache)
+except ImportError:
+    def _needs_psims(name):
+        raise ImportError("Loading %s requires the `psims` library. To access it, please install `psims" % name)
```
levitsky, line 72:
```suggestion
        raise ImportError("Loading %s requires the `psims` library. To access it, please install `psims`" % name)
```

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+    # Python 2 doesn't have a builtin Enum type
+    Enum = object
+
+
+from pyteomics import parser
+from pyteomics.mass import Composition, std_aa_mass, Unimod
+from pyteomics.auxiliary import PyteomicsError, BasicComposition
+from pyteomics.auxiliary.utils import add_metaclass
+
+# To eventually be implemented with pyteomics port?
```
levitsky, line 67: Is there anything you don't like about this dependency?

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+
+:py:func:`to_proforma` - Format a sequence and set of properties as ProForma text.
+
+
+Classes
+-------
+
+:py:class:`ProForma` - An object oriented version of the parsing and formatting code,
+coupled with minimal information about mass and position data.
+'''
```
levitsky, line 47:
```suggestion

Dependencies
------------

To resolve PSI-MOD, XL-MOD, and GNO identifiers, :mod:`psims` is required.

'''
```

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+try:
+    from psims.controlled_vocabulary.controlled_vocabulary import (load_psimod, load_xlmod, load_gno, obo_cache)
+except ImportError:
+    def _needs_psims(name):
+        raise ImportError("Loading %s requires the `psims` library. To access it, please install `psims" % name)
+
+    load_psimod = partial(_needs_psims, 'PSIMOD')
+    load_xlmod = partial(_needs_psims, 'XLMOD')
+    load_gno = partial(_needs_psims, 'GNO')
+    obo_cache = None
```
levitsky, line 77: This name does not seem to be used, is it necessary?

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+        }
+
+
+class XLMODResolver(ModificationResolver):
+    def __init__(self, **kwargs):
+        super(XLMODResolver, self).__init__('xlmod', **kwargs)
+        self._database = kwargs.get("database")
+
+    def load_database(self):
+        return load_psimod()
```
levitsky, line 399:
```suggestion
        return load_xlmod()
```

levitsky (commented):
This is huge, thank you!
I left some questions/comments in the code. I don't have any use cases, but I was able to catch a couple of issues by trying to parse examples from [ProForma README](https://github.com/HUPO-PSI/ProForma). We can try copying those into tests and adding a `psims` install to the GA workflow.
Otherwise, my only real concern is formula parsing. I left a comment about it in the code, too.
Thank you once again for the awesome work.

pyteomics/proforma.py
```diff
@@ -0,0 +1,1501 @@
...
+INTERVAL_TAG = ParserStateEnum.interval_tag
+TAG_AFTER = ParserStateEnum.tag_after_sequence
+POST_TAG_BEFORE = ParserStateEnum.post_tag_before
+UNLOCALIZED_COUNT = ParserStateEnum.unlocalized_count
+POST_GLOBAL = ParserStateEnum.post_global
+POST_GLOBAL_AA = ParserStateEnum.post_global_aa
+POST_INTERVAL_TAG = ParserStateEnum.post_interval_tag
+DONE = ParserStateEnum.done
+
+VALID_AA = set("QWERTYIPASDFGHKLCVNMX")
```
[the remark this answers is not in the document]
mobiusklein: You're right. The specification calls for U, O, B, Z, J, and X. I had overlooked the ambiguous ones and the two uncommon proteogenic amino acids. Fixed.

pyteomics/proforma.py
mobiusklein: Good catch and easily done. Thank you.

pyteomics/proforma.py
mobiusklein: The name gets baked into the `partial`-made function so that the error is clear about which "entity" depended upon the other source, e.g.
```python 
>>> load_psimod()
ImportError: Loading PSIMOD requires the `psims` library. To access it, please install `psims`
```
Technically, we could just make the message "Loading this controlled vocabulary requires `psims`." and be done, but it feels less explicit.

mobiusklein:
Thank you for catching those leftover items.

Compliance levels:

1) Base Level Support
Represents the lowest level of compliance, this level involves providing support for:
- [x] Amino acid sequences 
- [x] Protein modifications using two of the supported CVs/ontologies: Unimod and PSI-MOD. 
- [x] Protein modifications using delta masses (without prefixes)
- [x] N-terminal, C-terminal and labile modifications. 
- [x] Ambiguity in the modification position, including support for localisation scores.
- [x] INFO tag.

2) Additional Separate Support               
These features are independent from each other:
- [x] Unusual amino acids (O and U).
- [x] Ambiguous amino acids (e.g. X, B, Z). This would include support for sequence tags of known mass (using the character X).
- [x] Protein modifications using delta masses (using prefixes for the different CVs/ontologies).
- [x] Use of prefixes for Unimod (U:) and PSI-MOD (M:) names.
- [x] Support for the joint representation of experimental data and its interpretation.

3) Top Down Extensions
- [ ] Additional CV/ontologies for protein modifications: RESID (the prefix R MUST be used for RESID CV/ontology term names)
- [x] Chemical formulas (this feature occurs in two places in this list).

4) Cross-Linking Extensions
- [ ]  Cross-linked peptides (using the XL-MOD CV/ontology, the prefix X MUST be used for XL-MOD CV/ontology term names).

5) Glycan Extensions
- [x] Additional CV/ontologies for protein modifications: GNO (the prefix G MUST be used for GNO CV/ontology term names)
- [x] Glycan composition.
- [x] Chemical formulas (this feature occurs in two places in this list).

6) Spectral Support
- [ ] Charge and chimeric spectra are special cases (see Appendix II).
- [x] Global modifications (e.g., every C is C13).

pyteomics/proforma.py
levitsky: To be clear, this comment of mine referred to `obo_cache` only. It is not used in `pyteomics` code. Not sure why Github displayed four lines like that.

mobiusklein: Thank you for the clarifying. Yes, that variable is imported to allow `proforma` to expose control over the default file cache from `psims`. That way users could set the cache directory or disable the cache altogether if they wished. Otherwise they'd need to explicitly import it from `psims`. It probably isn't necessary to import it here and just needs to be documented clearly that it interacts with the `psims` cache mechanism.

pyteomics/proforma.py
```diff
@@ -0,0 +1,1580 @@
...
+            term = self.database[id]
+        else:
+            raise ValueError("Must provide one of `name` or `id`")
+        rec = {
+            "name":term.name,
+            "id": term.id,
+            "provider": self.name,
+            "composition": None,
+            "mass": self.get_mass_from_term(term)
+        }
```
levitsky, line 514: Should this method `return rec`?

pyteomics/proforma.py
```diff
@@ -0,0 +1,1580 @@
...
+
+    def __str__(self):
+        return to_proforma(self.sequence, **self.properties)
+
+    def __repr__(self):
+        return "{self.__class__.__name__}({self.sequence}, {self.properties})".format(self=self)
+
+    def __getitem__(self, i):
+        if isinstance(i, slice):
+            props = self.properties.copy()
```
levitsky, line 1483:
```suggestion

```

pyteomics/proforma.py
```diff
@@ -0,0 +1,1580 @@
...
+from functools import partial
+
+try:
+    from enum import Enum
+except ImportError:
+    # Python 2 doesn't have a builtin Enum type
+    Enum = object
+
+
+from pyteomics import parser
```
levitsky, line 112:
```suggestion
```

pyteomics/proforma.py
```diff
@@ -0,0 +1,1580 @@
...
+    - [ ] Charge and chimeric spectra are special cases.
+    - [x] Global modifications (e.g., every C is C13).
+
+
+'''
+
+from pyteomics.mass.mass import calculate_mass
+import re
+import warnings
+from collections import namedtuple, defaultdict, deque
```
levitsky, line 102:
```suggestion
from collections import deque
```

pyteomics/proforma.py
levitsky: Thank you! Indeed, I think it's worth mentioning in the docs because the user may not even know that `psims` is used, or how it works with caches.

levitsky (commented):
Thank you for solving the isotope parsing issue and laying out the supported features, this is very helpful. (Also for all the smaller edits.)
I noticed another couple of minor issues with the help of pyflakes.
Other than that, one question/suggestion I have is if you would agree with naming the entry level function just `parse`, for the sake of brevity and consistency (see `parser.parse` and all the `read` functions).

pyteomics/proforma.py
```diff
@@ -0,0 +1,1580 @@
...
+
+        Returns
+        reformatted : str
+            The re-written isotope notation
+        '''
+        parts = match.groupdict()
+        return "{element}[{isotope}]{quantity}".format(**parts)
+
+    def resolve(self):
+        normalized = ''.join(self.value.split(" "))
```
levitsky, line 631: Is this different from `self.value.replace(' ', '')`?

mobiusklein: Your way is more efficient. This is a left-over from my thinking about operating on each space-separated piece separately, but ultimately it wasn't necessary. I think the provision for spaces in formulae are an artefact of history now.

levitsky:
At this point I'm more than happy with the state of this PR. Please let me know if/when you think it's ready to merge.

mobiusklein:
Thank you.

There's another ProForma meeting tomorrow which may or may not introduce more changes. The ambiguous sequence region feature was a late addition. We'll see if more work is needed or if there are any comments from the group.

mobiusklein:
I've updated the documentation on `psims` to discuss the caching mechanism in a bit more detail. No new features have been added to ProForma since the last meeting, and likely the best way to get more feedback at this point is for people to try to use it. If you're satisfied with the level of documentation within the module itself, we can merge it.

Issue #40, opened 2021-05-06
mzML vs mzMLb vs mzXML


soerendip:
Hi, 

I made a quick benchmark with some metabolomics files and measured the read speed of various formats. For some reason I get the worst performance with mzMLb format. I wonder if you ever made a benchmark to compare the read speed in `pyteomics` with other programs. 

![read-speed-ms-file-formats](https://user-images.githubusercontent.com/3391614/117335369-5026b080-ae58-11eb-92e2-58f9aa6da383.png)

The orange bar is the bare read into memory. Without any conversion into a structured dataframe. _mzML_ and _mzXML_ was read with `pyteomics` while I used `pymzml` for _mzML_format.

mobiusklein:
When I was developing the mzMLb reader, we did do some benchmarking to show it was actually advantageous over mzML beyond the size reduction. It turned out to be sensitive to access patterns and the HDF5 compressor used. Are you able to share the files you benchmarked? I'd like to be able to look into where the slow-down you're observing is.

Were you sequentially reading scans from the file? The mzMLb reader does a small amount of optimization for sequential reading, more as a consequence of HDF5's chunked storage, but  this saves additional work like converting these chunks to Python objects. 

If you used the `psims` converter, did you also install `hdf5plugin`? One of the major advantages of mzMLb is that it opens the way for much faster compressors like `blosc` and `blosc:zstd` that are not part of the `h5py`/`hdf5` library  itself the way `zlib` is, but can be installed as plugins.  `zlib` has excellent compression ratios on a broad variety of data types, but it is quite slow. These newer compressors have been shown to be much faster without sacrificing much in terms of compression, particularly on numerical data. There's a pretty good chance that `blosc` will be required for the future. 

`mzXML` has much less metadata to decode, so it isn't surprising that might go faster.

soerendip:
The files were downloaded from `https://www.ebi.ac.uk/metabolights/MTBLS1569/descriptors`. And the converted files can be downloaded from `https://soerendip.com/dl/MTBLS1569/`

soerendip:
I installed some hdf5 library to make it work, I had opened a GitHub issue, but I don't find it, and I am not sure if it was related to pyteomics or psims.

mobiusklein:
[...]

PR #41, opened 2021-05-07
Merge read optimizations

mobiusklein:


levitsky:
Thank you. For future reference, this change is mentioned in #40.

PR #51, opened 2021-08-07
Fix ProForma Parsing and Update Docs

mobiusklein:
@bittremieux highlighted a few cases that were vague in the specification, and it turns out my parser didn't handle them correctly.

Also, recent changes to `psims` cause it to do more aggressive type coercion of CV terms, removing the need to do it in the `proforma` module itself. I've updated the minimum version required for the `proforma` feature to reflect this.

While there, I added a bit more documentation.


levitsky (approved):
Thank you!

mobiusklein:
These changes are ready to merge.

PR #54, opened 2021-08-17
Shuffling that preserves some amino acids

caetera:
Shuffle function in fasta module is updated to include two new argument
`keep_M` preserve **only** N-terminal methionine
`fix_aa` allows excluding some of amino acids from shuffling (i.e. position is preserved)


levitsky:
Thanks! I made a few minor changes, ready to accept if you have no objections.

caetera:
No objections whatsoever.

PR #61, opened 2021-10-15
Implement `charge_carrier` and `absolute` arguments for m/z calculation

### LINKED ISSUE #60: mass calculation with negative charge
Hi!

I've been using pyteomics for a while and it's great! Thanks a bunch!

I just noticed that attempting to calculate anion masses as below returns negative values. Is this the expected behavior? Not a big deal but I would have expected positive values so thought I'd mention it. This was with version 4.4.2.

All the best,
Michael

```
>>> from pyteomics import mass
>>> mass.calculate_mass(formula = 'C6H12O6', ion_type='M', charge = -1)
-179.05611163543
>>> mass.calculate_mass(formula = 'C6H12O6', ion_type='M', charge = -2)
-89.02441758433001
```

levitsky:
Closes #60. This is a suggested implementation of changes needed to support the discussed extensions. `absolute` is a new kwarg supported by `Composition.mass()` and `calculate_mass` transparently (this one is easy).

New `charge_carrier` argument, as well as `charge`, are now handled differently than `charge` used to be handled. They can stick to the `Composition` object or be passed to `Composition.mass()` (`calculate_mass` implements the second route).
This creates quite a number of edge cases where checks and exceptions are needed; however, it allows reusing the same composition to calculate m/z of different charge states and/or with different charge carriers more easily.

Objections and critique are welcome. Is this approach (keeping charge and charge carrier as attributes and accepting them as kwargs for mass calculation) too complex and error-prone, or is it worth the convenience?

I will be following up with more test coverage for new behavior, but a quick demo for now:

```
In [4]: mass.calculate_mass(formula = 'C6H12O6', charge = -1, charge_carrier='Cl-')
Out[4]: -215.0322407822

In [5]: mass.calculate_mass(formula = 'C6H12O6', charge = -1, charge_carrier='Cl-', absolute=True)
Out[5]: 215.0322407822

In [6]: mass.calculate_mass(formula = 'C6H12O6', charge = -1, charge_carrier='HCOO-', absolute=True)
Out[6]: 225.06104237339002
```


mobiusklein:
I'm inclined to say that everything is kept simpler by keeping charge and the charge carrier independent of the `Composition` type, as in your first proposal. As discussed in the issue, you can avoid complicating the infrastructure by instead treating the charge carrier as neutral (add or subtract however many charges' worth of protons to it) and then use the mass of a proton as the charge carrier and get the right answer. Also, less book keeping in all the copy/state saving methods.

This current approach also can't handle when you have a mix of charge carriers, e.g. two protons and a sodium cation, or worse, an iron cation adduct which may induce a +2 or +3 and a commensurate different change in mass and m/z (obligatory adduct calculator link: https://fiehnlab.ucdavis.edu/staff/kind/metabolomics/ms-adduct-calculator/). I've got a bunch in a single spectrum on the same molecule (4+ glycopeptide with four protons, three protons and ammonium, three protons and sodium, three protons and potassium, and an unknown number of protons and iron):

image

Does `std_ion_comp`  have anything to do with charge? All the values are neutral losses, except perhaps the fancy c and z ions, but my lack of knowledge of the ExD intermediate radicals and the dissociation mechanism makes it hard for me to say. I think they're called "ion" compositions because we don't regularly refer to "b" and "y" fragments outside of their charged forms.

Currently, the `Composition.mass` implementation has to do extra book keeping when you pass it a charge and the composition already has `H+` in it, which is extra work you may not want anyway. However, if you specify `ion_comp` and `ion_type` in the `Composition.__init__`, there is no "undo and overwrite" behavior in `Composition.mass`. I don't think we need to do away with these, but instead add an explicit `charge_carrier` param to `Composition.mass`, and have it determine:
1. Do I have a charge and a charge carrier?
    - Do I already have an H+ count?
           - Yes: Throw an error or try to figure out how to fix
           - No: Directly compute correct m/z from neutral composition mass plus charge times charge carrier mass divided by abs(charge)
2. Do I have an explicit charge carrier but no charge? If so, throw an error or assume singly charged
3. Do I have an explicit charge but no charge carrier? If so, assume charge carrier is a proton and assume that the composition is neutral or already contains an `H+` count and neutralize it before calculating m/z, current behavior but with abs(charge) in denominator instead of plain charge
4. Do I have no charge related parameters? Do current behavior but with abs(charge) in denominator instead of plain charge.

If you're worried about breaking code that sign-fixes after the fact, I'd add the `absolute` parameter, and check whenever charge is negative and emit a warning about this behavior changing in a future version.

levitsky:
I'm not sure I understand the benefit of this approach. What does this "neutral charged" composition represent?
It seems to me that by subtracting hydrogens we get a composition that does not correspond to a real molecule.
Since `Composition` can be used for things other than mass calculation, I expected that we'd want it to represent a specific substance/peptide in either charged or neutral form. In the neutral case, I thought we'd just build a composition without any knowledge of the charge carrier, then in `mass()` we'd get the charge and charge carrier information and apply them. What am I missing?

levitsky:
Rewritten in accordance with the discussion above. Charge-related parameters are only accepted in `mass()`. Charge carrier can be specified as a formula or composition. The group's own charge can be specified separately or inside the string spec. The implementation only works when all charges are homogeneous (no different charge carriers), so the total charge must be a multiple of the adduct's charge.

P.S. The test failure seems unrelated and comes from `usi`.

mobiusklein:
It looks like www.peptideatlas.org has either dropped that dataset or generally is re-arranging its website. Changing the backend seems to resolve the issue.

```patch
diff --git a/tests/test_usi.py b/tests/test_usi.py
index c3a8391..367e4ed 100644
--- a/tests/test_usi.py
+++ b/tests/test_usi.py
@@ -25,7 +25,7 @@ class USITest(unittest.TestCase):
 class PROXITest(unittest.TestCase):
     def test_request(self):
         usi_str = "mzspec:MSV000085202:210320_SARS_CoV_2_T:scan:131256"
-        response = proxi(usi_str, backend='peptide_atlas')
+        response = proxi(usi_str, backend='massive')
 
         assert set(usi_proxi_data.keys()) <= set(response.keys())
 ```

PR #67, opened 2021-12-17
Explicitly close the thread pool in PROXIAggregator on GC

### LINKED ISSUE #66: ThreadPool in PROXIAggregator not closed
The `PROXIAggregator` uses a `ThreadPool` to query multiple PROXI resources simultaneously. This `ThreadPool` is [stored between multiple function calls](https://github.com/levitsky/pyteomics/blob/master/pyteomics/usi.py#L304), I assume to avoid the overhead of creating the `ThreadPool`. However, it is never properly closed, resulting in the following error:

```
Exception ignored in: <function Pool.__del__ at 0x7fe660d174c0>
Traceback (most recent call last):
File "/home/wout/.conda/envs/spectrum_utils/lib/python3.9/multiprocessing/pool.py", line 268, in __del__
File "/home/wout/.conda/envs/spectrum_utils/lib/python3.9/multiprocessing/queues.py", line 372, in put
AttributeError: 'NoneType' object has no attribute 'dumps'
```

This does not interfere with my code, as it only happens at the end, presumably when all objects are cleaned up. However, it would be nice to avoid having this error at all.

Maybe the `ThreadPool` should just be created for resolving of individual USIs, and then immediately cleaned up? The overhead of repeatedly creating the pool should be pretty minimal, especially compared to REST API calls.

mobiusklein:
Closes #66

### DIFF
--- a/pyteomics/usi.py
+++ b/pyteomics/usi.py
@@ -282,12 +282,13 @@ class PROXIAggregator(object):
         to the number of servers to query.
     timeout : float
         The number of seconds to wait for a response.
-
+    ephemeral_pool : bool
+        Whether or not to tear down the thread pool between requests.
     '''
 
     _coalesce_resolution_methods = ("first", )
 
-    def __init__(self, backends=None, n_threads=None, timeout=15, merge=True, **kwargs):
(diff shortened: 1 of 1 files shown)


bittremieux (approved):

levitsky:
Thank you!

PR #69, opened 2022-01-17
Fix generic modification resolver

### LINKED ISSUE #68: [proforma] Unresolvable modification results in non-descriptive error when accessing `.mass` property
Hi @levitsky and @mobiusklein,

As the title says, in the `proforma` module, when trying to access the `.mass` property on an unresolvable modification, the following error is thrown:

```python
parsed_sequence, modifiers = proforma.parse("D[TMT6plex]PEPTIDK/2")
modification = parsed_sequence[0][1][0]
modification.mass

TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_767612/1789684910.py in <module>
      1 parsed_sequence, modifiers = proforma.parse("D[TMT6plex]PEPTIDK/2")
      2 modification = parsed_sequence[0][1][0]
----> 3 modification.mass

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in mass(self)
    687         -------float
    688         '''
--> 689         return self.definition['mass']
    690 
    691     @property

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in definition(self)
    677         '''
    678         if self._definition is None:
--> 679             self._definition = self.resolve()
    680         return self._definition
    681 

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in resolve(self)
    932         '''
    933         keys = self._parse_identifier()
--> 934         defn = self.resolver(*keys)
    935         if defn is not None:
    936             return defn

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in __call__(self, name, id, **kwargs)
    380 
    381     def __call__(self, name=None, id=None, **kwargs):
--> 382         return self.resolve(name, id, **kwargs)
    383 
    384     def __eq__(self, other):

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in resolve(self, name, id, **kwargs)
    617         for resolver in self.resolvers:
    618             try:
--> 619                 defn = resolver(name=name, id=id, **kwargs)
    620             except (KeyError):
    621                 continue

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in __call__(self, name, id, **kwargs)
    380 
    381     def __call__(self, name=None, id=None, **kwargs):
--> 382         return self.resolve(name, id, **kwargs)
    383 
    384     def __eq__(self, other):

~/anaconda3/envs/py37/lib/python3.7/site-packages/pyteomics/proforma.py in resolve(self, name, id, **kwargs)
    449         else:
    450             raise ValueError("Must provide one of `name` or `id`")
--> 451         mass = float(defn.DiffMono)
    452         diff_formula = str(defn.DiffFormula.strip().replace(" ", ''))
    453         composition = Composition(diff_formula)

TypeError: float() argument must be a string or a number, not 'NoneType'
```

Adding the Unimod prefix `U:` fixes the issue, in this case:

```python
parsed_sequence, modifiers = proforma.parse("D[U:TMT6plex]PEPTIDK/2")
modification = parsed_sequence[0][1][0]
modification.mass

>>> 229.162932
```

It is of course expected that the mass of an unresolvable modification cannot be retrieved, although I would propose to raise a more informative exception.

As a side question: How come that `Oxidation` is parsed without prefix, while `TMT6plex` is not?

Thanks!
Ralf

mobiusklein:
Fixes #68

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -187,6 +187,14 @@ class TagTypeEnum(Enum):
 _sentinel = object()
 
 
+class ModificationMassNotFoundError(ProFormaError):
+    pass
+
+
+class UnknownMonosaccharideError(ProFormaError):
+    pass
+
+
 @add_metaclass(PrefixSavingMeta)
 class TagBase(object):
     '''A base class for all tag types.
@@ -448,7 +456,10 @@ class PSIModResolver(ModificationResolver):
             defn = self.database['MOD:{:05d}'.format(id)]
         else:
             raise ValueError("Must provide one of `name` or `id`")
-        mass = float(defn.DiffMono)
+        try:
+            mass = float(defn.DiffMono)
+        except (KeyError, TypeError, ValueError):
+            raise ModificationMassNotFoundError("Could not resolve the mass of %r from %r" % ((name, id), defn))
         diff_formula = str(defn.DiffFormula.strip().replace(" ", ''))
         composition = Composition(diff_formula)
         return {
@@ -475,7 +486,10 @@ class XLMODResolver(ModificationResolver):
             defn = self.database['XLMOD:{:05d}'.format(id)]
         else:
             raise ValueError("Must provide one of `name` or `id`")
-        mass = float(defn['monoIsotopicMass'])
+        try:
+            mass = float(defn['monoIsotopicMass'])
+        except (KeyError, TypeError, ValueError):
+            raise ModificationMassNotFoundError("Could not resolve the mass of %r from %r" % ((name, id), defn))
         if 'deadEndFormula' in defn:
             composition = Composition(defn['deadEndFormula'].replace(" ", '').replace("D", "H[2]"))
         elif 'bridgeFormula' in defn:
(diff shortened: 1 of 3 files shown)


pyteomics/proforma.py
```diff
@@ -448,7 +456,10 @@ def resolve(self, name=None, id=None, **kwargs):
             defn = self.database['MOD:{:05d}'.format(id)]
         else:
             raise ValueError("Must provide one of `name` or `id`")
-        mass = float(defn.DiffMono)
+        try:
+            mass = float(defn.DiffMono)
```
levitsky, line 460: Thank you!
It appears that on Python 2 the new test fails because the expected exception is not raised. Where on Python 3 `defn.DiffMono` is `None` here, on Python 2 it is equal to `229.162932` for `"TMT6plex"`. I'm not sure of the exact reason yet, maybe you see what is happening?

mobiusklein:
The reason why that failed on Py2 is that the `dict` ordering is different than on Py3, and `TMT6plex` is a registered synonym for every TMT-labeled residue _and_ reporter ion in PSI-MOD, which means that it is legal (semantically) to resolve to any of them. Therefore, `MOD:TMT6plex` is unsafe to use as its behavior is non-deterministic on Py2.

I changed that test to use an explicit ID to avoid this type of error in the tests.

PR #70, opened 2022-02-13
Switch USI test backend

mobiusklein:
This patch fixes the USI testing backend, it seems like the Peptide Atlas USI service is down, while the MassIVE service is still working. For all example USIs at http://proteomecentral.proteomexchange.org/usi/, MassIVE works while PeptideAtlas does not.


levitsky:
Thank you!

Issue #82, opened 2022-10-21
[Pitch] Include unimod with the release of pyteomics


jspaezp:
I have noticed that when using the unimod interface, the default is to download the database.
I would like to propose to include the database with the release.

It would greatly improve the speed in using the functionality.

let me know what you think!
best
Sebastian

mobiusklein:
Not that it is a bad thing to do (I do it in other libraries, specifically with Unimod and other controlled vocabularies in `psims`), but both the in-memory and SQLite3-based Unimod interfaces let you pass the `Unimod` type object at a path or file-like object containing the database's XML representation. If load-time is a concern, you can store a copy of the database locally and/or packaged with your other code.

In the case of the SQLite3-based interface, this is encouraged since it's all for the persistent on-disk representation.

levitsky:
I'd like to hear opinions: do we need to better document the possibility of using a local copy of Unimod, or is it not convenient enough?

jspaezp:
Going back to the docs to see what I missed I realized 2 main things.

I missed the documentation where passing a path is possible to the unimod class instantiation. (in my defense, it is really deep between a lot or other text)
here: https://pyteomics.readthedocs.io/en/latest/api/unimod.html#pyteomics.mass.unimod.Unimod

On the other hand, I was using it as part of the proforma parser, which uses the `UnimodResolver` object, which internally loads an `Unimod` object that does not take any arguments.
https://github.com/levitsky/pyteomics/blob/608ae01e111093b36f8ab22c8d0aefcb74afc01a/pyteomics/proforma.py#L401-L408

so ... Would I need to subclass the resolver, over-write the `load_database` method so the instance of the unimod object reads from disk? (this feels overly complicated).

On the other hand, I am having a hard time thinking of an use case where one would not want the database to be available in disk instead of downloading it in every instantiation.

So I guess, ... to summarize it ...

1. what is the rationale behind having it downloaded instead of bundled with the python package?
2. Docs could be improved by moving the documentation on the __init__(self, path) section of the Unimod class to a more prevalent location (maybe even add an example), lmk if you would like some help with that.
3. the proforma modification resolver does not have that option of pointing to a disk location. (feature request??)

let me know what you think!
as always thank you very much for the help and reply!

mobiusklein:
Yes, the `ProForma` interface doesn't support overriding the location. It uses `psims` for everything but `Unimod`. The workaround currently would be to do something like this:
```python
from pyteomics import mass, proforma

[...]

PR #85, opened 2022-10-25
Improve documentation of Unimod loading and highlight in ProForma

mobiusklein:
This PR implements some of the changes discussed in #82 but does not outright bundle the CV source file with the codebase. It adds documentation about how the CV can be loaded from disk instead of the internet.

It also re-structures the `pyteomics.proforma` documentation a bit and in need of a bit meditation I wrote a `fragments` method for the `Proforma` type.

### DIFF
--- a/doc/source/api/proforma.rst
+++ b/doc/source/api/proforma.rst
@@ -11,6 +11,9 @@ for more up-to-date information.
 
 Strictly speaking, this implementation supports ProForma v2.
 
+.. contents::
+    :depth: 2
+
 Data Access
 -----------
 
--- a/doc/source/api/proforma.rst
+++ b/doc/source/api/proforma.rst
@@ -1,1 +1,229 @@
-.. automodule:: pyteomics.proforma
+
+.. module:: pyteomics.proforma
+
+proforma - Proteoform and Peptidoform Notation
+==============================================
+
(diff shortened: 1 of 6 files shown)


mobiusklein:
I copied the docstrings because I cannot figure out how to tell `autodoc`'s `automodule` directive to just copy the module docstring and not list out the module members too. I couldn't make `:no-members:` or limiting it to only a subset of members. In order for the classes to be laid out in a way that isn't overwhelming to the reader.

I needed to disable `automodule` so I could intersperse them with prose to explain _why_ there are all these classes but the only thing they need to really worry about is the functional API or the `ProForma` object.

levitsky:
I feel like a lot of other documentation should be reworked for readability as well (outside of this PR, of course).

Do you think the module docstrings should be kept still? My concern is, the two versions will diverge; also I am not sure if anyone really benefits from the module-level documentation being there (unlike members' documentation, which often helps in interactive environments).

mobiusklein:
I removed the complicated module docstring content. I declined to put an actual URL to the docs in the docstring because I didn't know if you have persistent URL set up.

While I was at it I added a test for the new `fragments` method and made it 2.5x faster.

levitsky (approved):

levitsky:
Thank you!

PR #95, opened 2022-12-02
Resolve `referenceableParamGroupRef`

mobiusklein:
This change adds another special case to `XML._get_info` to treat `referenceableParamGroupRef` similar to how we treat `cvParam` and `userParam`, and then hands over responsibility for _how_ that special handling is defined to the implementing subclass, so long as the result is a `List[xml._XMLParam]`.

The implementation for mzML defers parsing `referenceableParamGroup` until you encounter a `referenceableParamGroupRef` and then does a random access seek from the start of the file parsing 

Reasons why this might not be desirable:
1. When you are reading an stream you cannot call `seek` on (e.g. `stdin` or a `socket`), you can't jump back to the start of the file to read the param group definitions.
2. Right now, `iterfind` re-scans the _whole file_. We need to add a method to halt parsing after we get to a certain element.

As noted in the discussion, this behavior is irrelevant when used with `retrieve_refs=True`, but that is not on by default.

### DIFF
--- a/pyteomics/mzml.py
+++ b/pyteomics/mzml.py
@@ -129,6 +129,7 @@ class MzML(xml.ArrayConversionMixin, aux.TimeOrderedIndexedReaderMixin, xml.Mult
 
     def __init__(self, *args, **kwargs):
         self.decode_binary = kwargs.pop('decode_binary', True)
+        self._referenceable_param_groups = None
         super(MzML, self).__init__(*args, **kwargs)
 
     def __getstate__(self):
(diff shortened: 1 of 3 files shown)


mobiusklein:
Draft of early halting iterparse:
```python
@xml._keepstate
def iterparse_until(source, target_name, quit_name):
    '''Iteratively parse XML stream in ``source``, yielding XML elements
    matching ``target_name``. If at any point a tag matching ``quit_name``
    is encountered, stop parsing.

    Parameters
    ----------
    source: file-like
        A file-like object over an XML document
    tag_name: str
        The name of the XML tag to parse until
    quit_name: str
        The name to stop parsing at.

    Yields
    ------
    lxml.etree.Element
    '''
    g = etree.iterparse(source, ('start', 'end'))
    for event, tag in g:
        if event == 'start':
            if xml._local_name(tag) == quit_name:
                break
            else:
                if xml._local_name(tag) == target_name:
                    yield tag
                else:
                    tag.clear()
```

mobiusklein:
Rewrote how param groups are read on-demand to avoid doing a full scan of the file _unless_ there is no `referenceableParamGroup` with the requested `id`. That simplifies things.

levitsky:
Thank you!
Do you think it's beneficial to index those refs? That way `get_by_id` is supposed to seek immediately to the right place instead of a sequential scan. If those param groups are in the beginning, the savings could be negligible though.

Or am I missing something?

mobiusklein:
The param groups are always near the top of the file, so it's of limited value to index them.

PR #99, opened 2023-01-31
Fix issue #98 for colon-containing mods

### LINKED ISSUE #98: ProForma parser can't handle a modification name with colon if it occurs without ontology prefix
Some modifications, including cation adducts, have a colon in their name, e.g. [`Cation:Na`](http://www.unimod.org/modifications_view.php?editid1=30).

`pyteomics.proforma` can't handle those when the ontology prefix is not specified:

```python
In [18]: proforma.ProForma.parse('VKATAGDTHLGGED[Cation:Na]FDNRLVNHLATE')
KeyError                                  Traceback (most recent call last)
Input In [18], in <cell line: 1>()
----> 1 proforma.ProForma.parse('VKATAGDTHLGGED[Cation:Na]FDNRLVNHLATE')

File ~/py/pyteomics/pyteomics/proforma.py:2037, in ProForma.parse(cls, string)
   2024 @classmethod
   2025 def parse(cls, string):
   2026     '''Parse a ProForma string.
   2027 
   2028     Parameters
   (...)
   2035     ProForma
   2036     '''
-> 2037     return cls(*parse(string))

File ~/py/pyteomics/pyteomics/proforma.py:1592, in parse(sequence)
   1590 if c in VALID_AA:
   1591     if current_aa is not None:
-> 1592         positions.append((current_aa, current_tag() if current_tag else None))
   1593     current_aa = c
   1594 elif c == '[':

File ~/py/pyteomics/pyteomics/proforma.py:1408, in TokenBuffer.__call__(self)
   1407 def __call__(self):
-> 1408     return self.process()

File ~/py/pyteomics/pyteomics/proforma.py:1457, in TagParser.process(self)
   1456 def process(self):
-> 1457     value = super(TagParser, self).process()
   1458     if not isinstance(value, list):
   1459         value = [value]

File ~/py/pyteomics/pyteomics/proforma.py:1398, in TokenBuffer.process(self)
   1396     value = [self._transform(v) for v in self.tokenize()]
   1397 else:
-> 1398     value = self._transform(self.buffer)
   1399 self.reset()
   1400 return value

File ~/py/pyteomics/pyteomics/proforma.py:1451, in TagParser._transform(self, value)
   1450 def _transform(self, value):
-> 1451     tag = process_tag_tokens(value)
   1452     if tag.group_id:
   1453         self.group_ids.add(tag.group_id)

File ~/py/pyteomics/pyteomics/proforma.py:1106, in process_tag_tokens(tokens)
   1104         main_tag = GenericModification(''.join(value))
   1105     else:
-> 1106         tag_type = TagBase.find_by_tag(prefix)
   1107         main_tag = tag_type(value)
   1108 if len(parts) > 1:

File ~/py/pyteomics/pyteomics/proforma.py:82, in PrefixSavingMeta.find_by_tag(self, tag_name)
     80     raise ValueError("tag_name cannot be None!")
     81 tag_name = tag_name.lower()
---> 82 return self.prefix_map[tag_name]
```

This works fine:

```python
proforma.ProForma.parse('VKATAGDTHLGGED[UNIMOD:Cation:Na]FDNRLVNHLATE')
```

The [spec](https://github.com/HUPO-PSI/ProForma/blob/master/SpecDocument/Release_v2.0/ProForma_v2_Final.pdf) says that the Unimod prefix is optional, which does work with other names.

mobiusklein:
Closes #98 

The behavior is now to treat any colon-containing tag that doesn't have a known prefix as though it were a `GenericModification` tag. This will lead to strange behavior if we encounter a tag type that we don't support like `RESID` or `XLMOD`, with these tags being wrapped in `GenericModification` instances and then having them explode when someone tries to resolve their properties.

I can add an "unsupported list" to explicitly check those tag types and throw an error pre-emptively, but that might prevent a reasonable use-case of using `pyteomics.proforma` as a path-through between the user and some other service.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+4.5.7a4
+-------
+ - Fix issue #98 by Joshua Klein
+
 4.5.7a3
 -------
 
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -1103,8 +1103,16 @@ def process_tag_tokens(tokens):
         if prefix is None:
             main_tag = GenericModification(''.join(value))
         else:
-            tag_type = TagBase.find_by_tag(prefix)
-            main_tag = tag_type(value)
+            try:
+                tag_type = TagBase.find_by_tag(prefix)
+                main_tag = tag_type(value)
+            except KeyError:
+                main_tag_str = ''.join(main_tag)
+                warnings.warn(
+                    f"Possible unmatched prefix detected for {main_tag_str},"
+                     " attempting to resolve as generic modification")
+                main_tag = GenericModification(main_tag_str)
+
     if len(parts) > 1:
         extras = []
         for part in parts[1:]:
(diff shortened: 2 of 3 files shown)


levitsky:
Thank you @mobiusklein!
cc: @caetera

PR #101, opened 2023-02-03
Address broader ProForma colon ambiguity

mobiusklein:
Fixes #99

This refactors how `:` characters are treated more broadly to address the issue in #99, moving the responsibility to the modification resolver, not the modification tag. It is an extension of #98.


levitsky (approved):
Looks good to me, thank you!

PR #105, opened 2023-02-25
Pass through the decode_binary parameter in read for mzml and mzxml

### LINKED ISSUE #104: [mzml] `decode_binary` argument documented but not implemented
Hi everyone,

In the mzml reader documentation for `read`, the `decode_binary` argument is listed, but it does not seem to be implemented. Setting the argument results in a TypeError:
```
TypeError: read() got an unexpected keyword argument 'decode_binary'
```

Documentation:
https://github.com/levitsky/pyteomics/blob/7c7971d468394a1844059a0af605111e515c25b2/pyteomics/mzml.py#L387-L390

Function definition:
https://github.com/levitsky/pyteomics/blob/7c7971d468394a1844059a0af605111e515c25b2/pyteomics/mzml.py#L360

I guess the documentation is a left-over from previous functionality?

Best,
Ralf

mobiusklein:
Fixes #104

Similar problem found for `mzxml` which is fixed as well. `mzmlb` does not this option in its `read` function.

### DIFF
--- a/pyteomics/mzml.py
+++ b/pyteomics/mzml.py
@@ -357,7 +357,7 @@ class MzML(xml.ArrayConversionMixin, aux.TimeOrderedIndexedReaderMixin, xml.Mult
         return scan['scanList']['scan'][0]['scan start time']
 
 
-def read(source, read_schema=False, iterative=True, use_index=False, dtype=None, huge_tree=False):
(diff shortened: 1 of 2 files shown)


levitsky:
Thank you!

PR #120, opened 2023-09-01
Add raw fasta file header to parsers

caetera:
This PR adds an entry to FASTA file parsers that contains raw (i.e. non-parsed header). The latter is used by the `write` to write parsed protein entries.

Useful, for the fast filtering of FASTA files, for example:
```
from pyteomics import fasta
fasta_entries = fasta.UniProt('uniprot_multi_organism.fasta')
fasta.write([protein for protein in fasta_entries if protein.description.get('OX', -1) == 123], 'uniprot_single_organism.fasta')
```

### DIFF
--- a/pyteomics/fasta.py
+++ b/pyteomics/fasta.py
@@ -117,11 +117,12 @@ from . import auxiliary as aux
 
 Protein = namedtuple('Protein', ('description', 'sequence'))
 DECOY_PREFIX = 'DECOY_'
+RAW_HEADER_KEY = '__raw__'
 
 
 def _add_raw_field(parser):
     """
-    Add "_raw" field to the parsed dictinary
+    Add :py:const:`RAW_HEADER_KEY` field to the parsed dictinary.
 
     Parameters
     ----------
@@ -133,13 +134,13 @@ def _add_raw_field(parser):
     None.
 
     """
-    def _new_parser(cls, descr):
-        parsed = parser(cls, descr)
-        if not '_raw' in parsed.keys():
-            parsed['_raw'] = descr
+    def _new_parser(instance, descr):
+        parsed = parser(instance, descr)
+        if not RAW_HEADER_KEY in parsed:
+            parsed[RAW_HEADER_KEY] = descr
         else:
-            raise aux.PytemicsError('Cannot save raw protein header, since the corresponsing\
-                                    key (_raw) already exists')
+            raise aux.PytemicsError('Cannot save raw protein header, since the corresponsing'
+                                    'key ({}) already exists.'.format(RAW_HEADER_KEY))
         return parsed
 
     return _new_parser
@@ -655,8 +656,8 @@ def write(entries, output=None):
     ----------
     entries : iterable of (str/dict, str) tuples
         An iterable of 2-tuples in the form (description, sequence).
-        If description is a dictionary, it should contain a key "_raw"
-        that will be used for protein description.
+        If description is a dictionary, the value for :py:const:`RAW_HEADER_KEY`
+        will be written as protein description.
     output : file-like or str, optional
         A file open for writing or a path to write to. If the file exists,
         it will be opened for writing. Default is :py:const:`None`, which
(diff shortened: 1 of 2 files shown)


levitsky:
Thank you, this looks great. I made a couple of minor changes, if you don't mind.

I also looked into adding this via a metaclass so that this is injected in every `FlavoredMixin` subclass automatically, but do we have a universal way to add a proper metaclass? It looks like the `add_subclass` routine actually modifies the class itself and does not affect its descendants, unless I'm missing something, which is totally possible since I don't deal with metaclasses often.

mobiusklein:
@levitsky I'm not sure what the `add_subclass` name you're referring to is. By "universal way to add a proper metaclass", do you mean syntactically to deal with the differences between Py2 and Py3? There's `six.add_metaclass`, which produces functionally the same thing:

```mermaid
flowchart types
  A[class FooMeta]
```

levitsky:
@mobiusklein thanks for chiming in. I mean the function in `auxiliary.utils`, and yes, I was referring to a way to make it work on Py2 and Py3. We've been avoiding adding `six` as a dependency and I thought `utils.add_subclass` was functionally equivalent?

mobiusklein:
Yes, `pyteomics.auxiliary.utils.add_metaclass` is functionally equivalent to `six.add_metaclass`. I forgot I added that (probably when I wrote the `mztab` module). It does affect subclasses, because the object bound to the name of the class you decorate isn't the `type` you wrote out syntactically. `add_metaclass` creates a copy of that type object's internal data structures, and then creates a new type object that is an instance of the `metaclass` instead of the built-in `type`.

So to accomplish what you're trying to do with `_add_raw_field` with a metaclass, you'd write something like this:

```python
class _FastaParserFlavorMeta(type):
    def __new__(mcs, name, bases, namespace):
        if "parse" in namespace:
            namespace["parse"] = _add_raw_field(namespace["parse"])
        return super().__new__(mcs, name, bases, namespace)

@add_metaclass(_FastaParserFlavorMeta)
class FlavorMixin(...):
    ...
```
Here, `namespace` is a `dict` of all the names bound to the class definition, e.g. methods, class attributes, annotations, docstring, and a few other things. If the class defines a `parse` method, it'll be in the `namespace`. Otherwise, that'll be resolved by the usual MRO from `bases`. If every class you create that defines a `parse` method is an instance of `_FastaParserFlavorMeta`, they'll always have this wrapped in `_add_raw_field`. If you were to use a mixin type that is not a `_FastaParserFlavorMeta` to inject your parse method won't be wrapped unless you explicitly write a forwarding method like so: 

```python
class ExternalParseMixin:
    def parse(self, *args, **kwargs):
        ...

# Inherit `_FastaParserFlavorMeta` from `FlavorMixin`, 
class NewFlavorMixin(FlavorMixin, ExternalParseMixin):
    def parse(self, *args, **kwargs):
        return ExternalParseMixin.parse(self, *args, **kwargs)
```
There are a few other things we did to the FASTA parsers over the years we might be able to make less repetitive with a metaclass, like `_add_init`, but I'd need to re-read it to remember what problem was.

mobiusklein:
Yes, `pyteomics.auxiliary.utils.add_metaclass` is functionally equivalent to `six.add_metaclass`. I forgot I added that (probably when I wrote the `mztab` module). It does affect subclasses, because the object bound to the name of the class you decorate isn't the `type` you wrote out syntactically. `add_metaclass` creates a copy of that type object's internal data structures, and then creates a new type object that is an instance of the `metaclass` instead of the built-in `type`.

So to accomplish what you're trying to do with `_add_raw_field` with a metaclass, you'd write something like this:

```python
class _FastaParserFlavorMeta(type):
    def __new__(mcs, name, bases, namespace):
        if "parse" in namespace:
            namespace["parse"] = _add_raw_field(namespace["parse"])
        return super().__new__(mcs, name, bases, namespace)

@add_metaclass(_FastaParserFlavorMeta)
class FlavorMixin(...):
    ...
```
Here, `namespace` is a `dict` of all the names bound to the class definition, e.g. methods, class attributes, annotations, docstring, and a few other things. If the class defines a `parse` method, it'll be in the `namespace`. Otherwise, that'll be resolved by the usual MRO from `bases`. If every class you create that defines a `parse` method is an instance of `_FastaParserFlavorMeta`, they'll always have this wrapped in `_add_raw_field`. If you were to use a mixin type that is not a `_FastaParserFlavorMeta` to inject your parse method won't be wrapped unless you explicitly write a forwarding method like so: 

```python
class ExternalParseMixin:
    def parse(self, *args, **kwargs):
        ...

# Inherit `_FastaParserFlavorMeta` from `FlavorMixin`, 
class NewFlavorMixin(FlavorMixin, ExternalParseMixin):
    def parse(self, *args, **kwargs):
        return ExternalParseMixin.parse(self, *args, **kwargs)
```

mobiusklein:
It's safe to derive `_FastaParserFlavorMeta` from `ABCMeta`, with the caveat that if you decorate `parser` with `abstractmethod`, I think the current trick of patching the `namespace` will cause trouble, though this can be countered by only running that code when the type is _not_ the base `FlavorMixin`.

Since that's not currently the design, that shouldn't be an issue, although semantically `parser` _is_ treated like an abstract method, we just don't enforce it, nor do we need to make things too much more complicated.

levitsky:
Thank you all! Looks ready to me.

PR #137, opened 2023-12-28
Calculate m/z for proforma in calculate_mass

levitsky:
This PR adds a new optional argument to `mass.calculate_mass`, `proforma`. If specified, it should be a ProForma string or a `proforma.ProForma` object. Its `mass` property is then converted to m/z if needed.

### DIFF
--- a/pyteomics/mass/mass.py
+++ b/pyteomics/mass/mass.py
@@ -367,6 +367,51 @@ class Composition(BasicComposition):
             charge = int(charge) * (-1, 1)[sign == '+']
         return comp, charge
 
+    @staticmethod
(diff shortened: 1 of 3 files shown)


mobiusklein:
This looks good to me, thank you for catching that duplicate water constant.

levitsky:
Thank you for checking it @mobiusklein !

levitsky:
Hi @FrangoST, thank you for reaching out!
This was not an intended change, I'm sorry for the inconvenience. I made a change to fix this in `master`, and if everything works as expected now, I will make a release shortly.

Issue #134, opened 2023-12-09
MGF File Loading Error - IndexedMGF Object Empty


CCranney:
Hi,

I am trying to load a metabolome library file in .mgf format using pyteomics. I've used it before on other library files, but in this instance the IndexedMGF object created by pyteomics is empty. For instance, where the following will generally produce a value greater than 0 for any appropriately-formatted and populated .mgf file:

```python
from pyteomics import mgf

lib = mgf.read('<path-to-file>')
print(len(lib))
```
[metabolome.tar.gz](https://github.com/levitsky/pyteomics/files/13621898/metabolome.tar.gz)

The above file is from GNPS. With this particular file it returns `0`, despite there being spectra in the file that appear to have the appropriate format. No errors are thrown, and I've spent the past little while trying to step through the code or just straight comparing a working example with this metabolome file example, but can't seem to find why it's not being read properly. Any chance you could take a look and see what the issue may be?

levitsky:
Hi @CCranney,

this MGF file doesn't have TITLEs for spectra, which is used as spectrum ID by default. A supported alternative is to use SCANS instead, which are present in this file, so this works:

```python
In [1]: from pyteomics import mgf

In [2]: f = mgf.IndexedMGF(filename, index_by_scans=True)

In [3]: len(f)
Out[3]: 24563
```

CCranney:
Works like a charm, thank you so much! Closing the issue.

CCranney:
I'm considering the issue closed, but this could be an enhancement request to throw a warning error if an otherwise-normal file has the same problem.

PR #138, opened 2024-01-09
Add a warning when an empty index is created, customize it for mgf

levitsky:
This addresses https://github.com/levitsky/pyteomics/issues/134#issuecomment-1850726407 by raising a warning if an `IndexedTextReader` creates an empty offset index. `IndexedMGF` has a customized warning text, additionally advising to toggle `index_by_scans`.

Would this be a welcome change?
A possible point of concern is that the warning is triggered regardless of "Index Saving", i.e. it doesn't matter if the index is built or just read from a file. Would we want to silence it if it's read from a file? Did I perhaps miss anything else?


levitsky:
After merging #142, I changed this PR so that the warning is triggered in `TextReaderMixin.build_byte_index`. If `IndexSavingMixin.build_byte_index` is successful in reading the index from the external file, the warning is not raised.

PR #142, opened 2024-02-01
Rename _build_index to build_byte_index

### LINKED ISSUE #140: IndexSavingMixin._build_index() is never called by text parsers?
Hi @mobiusklein, a quick question. I was looking at [`IndexSavingMixin`](https://github.com/levitsky/pyteomics/blob/master/pyteomics/auxiliary/file_helpers.py#L490) and [`IndexedTextReader`](https://github.com/levitsky/pyteomics/blob/master/pyteomics/auxiliary/file_helpers.py#L388) for ways to improve #138 and I noticed that the index saving functionality is implemented in `_build_index()`, which tries to read byte offsets from disk before falling back onto `super()._build_index()`. However, there is no such method in parent classes, as the index is built in `build_byte_index`, and so no method ever calls `IndexSavingMixin._build_index()`, either.

Is this perhaps an unintended result of some renaming in `IndexedTextReader` or am I misunderstanding something?

levitsky:
Fixes #140.

- unified method name is now `build_byte_index` across all indexed and index-saving parsers;
- the method returns the index rather than assign it (the behavior of `build_byte_index` is preserved);
- warning is thrown when defining an `IndexedXML` subclass with a `_build_index` method (Python 3 only).

@mobiusklein could you please take a look at it?

### DIFF
--- a/pyteomics/auxiliary/file_helpers.py
+++ b/pyteomics/auxiliary/file_helpers.py
@@ -535,25 +535,24 @@ class IndexSavingMixin(NoOpBaseReader):
             self._offset_index.save(f)
 
     @_keepstate_method
-    def _build_index(self):
+    def build_byte_index(self):
         """Build the byte offset index by either reading these offsets
         from the file at :attr:`_byte_offset_filename`, or falling back
-        to the method used by :class:`IndexedXML` if this operation fails
+        to the method used by :class:`IndexedXML` or :class:`IndexedTextReader` if this operation fails
         due to an IOError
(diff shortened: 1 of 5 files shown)


mobiusklein (approved):
There is a small change that I'll make in my downstream code to account for https://github.com/levitsky/pyteomics/pull/142/files#diff-7326fe871ea5482e4c336e02fabaa716957ab8aa12670f129c09d95d26630cceR538, but this looks good to me.

PR #158, opened 2024-09-28
Fix silencing of errors during ProForma mass calculations

### LINKED ISSUE #157: Unparsable modifications are silently ignored in ProForma mass calculation
Hello,

I discovered a quirky "attitude" of ProForma parsing.

If a ProForma object has a modification that cannot be decoded this modification is silently ignored in all further calculations, while the same calculation on the modification object raises. I believe it will be better to propagate the Exception rather than silencing it.

Here is an example (Pyteomics 4.7.2 Python 3.10.2 Win64)

Both raise `KeyError`

```python
proforma.GenericModification('Generic').mass

proforma.UnimodModification('123456').mass
```

However - this works while ignoring the modification

```python
for peptide in ['PEPTIDE', 'PEPTID[Generic]E', 'PEPTID[UNIMOD:123456]E']:
    prof = proforma.ProForma.parse(peptide)
    print(f'{prof}\n{prof.mass}\n')
```
```
PEPTIDE
799.3599640267099

PEPTID[Generic]E
799.3599640267099

PEPTID[UNIMOD:123456]E
799.3599640267099
```

mobiusklein:
Closes #157 

This issue also masked an error in one of the unit tests, where an attribute was misspelled. Previously, I was silently ignoring errors because not all tags have a mass value. Instead, now each tag is asked if it is expected to have a mass value, and if so, it is accessed and added to the accumulator.

Since modifications load lazily, and then read the mass value out of a `dict`, the `KeyError` could either be the result of a lookup failure for a modification resolver, or a lookup failure for the resolved properties of the modification, which would silence both mass lookup issues and modification resolution issues. Both of these are now resolved.

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -207,6 +207,16 @@ class TagBase(object):
     def parse(cls, buffer):
         return process_tag_tokens(buffer)
 
+    def has_mass(self):
+        """
+        Check if this tag carries a mass value.
+
+        Returns
+        -------
+        bool
+        """
+        return False
+
 
 class GroupLabelBase(TagBase):
     __slots__ = ()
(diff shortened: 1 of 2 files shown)


levitsky:
Thank you @mobiusklein!

PR #164, opened 2025-03-10
Update usi.py

dikshant182004:
just add a small change ,we can add a default value also ,but i thing filtering the pairs would be optimal approach

### DIFF
--- a/pyteomics/usi.py
+++ b/pyteomics/usi.py
@@ -371,7 +371,7 @@ class PROXIAggregator(object):
 
         def collapse_attribute(values):
             try:
-                acc = list(set(v['value'] for v in values))
+                acc = list(set(v['value'] for v in values if 'value' in v ))
             except TypeError:
                 acc = []
                 for v in values:


pyteomics/usi.py
```diff
@@ -371,7 +371,7 @@ def coalesce(self, responses, method='first'):
 
         def collapse_attribute(values):
             try:
-                acc = list(set(v['value'] for v in values))
+                acc = list(set(v['value'] for v in values if 'value' in v ))
```
mobiusklein, line 374: It makes more sense to use `list(set(v.get('value', '') for v in values if 'value' in v ))`, otherwise this will simply discard those attributes. If they don't have a value they might still indicate "something".

It looks like PRIDE serves units in this fashion, which unfortunately is not very useful as-is.

dikshant182004: so we can add a default value for them or mark them as None becz while testing spectrum utils our tests are being encountered with the keyerror associated with it .

dikshant182004: and this will also maintain the obtained data structure for future use and scalability

mobiusklein: Sorry, I meant this: `list(set(v.get('value', '') for v in values))`. That way `KeyError` never gets thrown, but we don't drop attributes without values.

dikshant182004: okay i got it now , this will work.

dikshant182004:
hi @mobiusklein ,i have updated this pr do have a look

dikshant182004:
hi @mobiusklein ,if u have a moment then plz have a look at the pr .

levitsky:
Since @mobiusklein approves and the workflows are running fine, I am going to merge this. Thank you both!

Issue #161, opened 2025-03-07
Should we use psims.obo to better control param value types in PSI formats?


levitsky:
I came across a corner case when parsing an MzIdentML file which has `spectrum title` filled in with bare scan numbers (integers):
```xml
<cvParam name="spectrum title" value="34" cvRef="PSI-MS" accession="MS:1000796" />
```

 The expected value type for this parameter would of course still be `str`, but the current logic in `XML._handle_param` only looks at the `type` attribute, which is not there:

https://github.com/levitsky/pyteomics/blob/24a2037665e929adace6f1491bd1ff1721362bf5/pyteomics/xml.py#L363-L369

Because float conversion succeeds in this case, `spectrum title` is returned as a float.

I don't know how many of such corner cases exist with other cvParams, but one way to fix this more or less systematically would be to invoke the PSI-MS ontology. Would it be wise to add `psims` as an extra dependency for `[XML]` and use `OBOCache` and `OBOParser` to look up terms there in `_handle_param` (at least when `type` attribute is not provided)?

@mobiusklein your input would be much appreciated. I feel like we could have discussed it at some point but I couldn't find it except for a discussion related to [unit awareness](https://levitsky.github.io/bitbucket_backup/#!/levitsky/pyteomics/issues/15/page/1).

mobiusklein:
There are an uncountable number of cases like this. All it takes is for a parameter whose value may be any string to be a string that happens to be a number.

When I wrote XSD type parsing into `psims`, I had this idea in mind, but I don't remember why I didn't go for it way back in the PR you linked. 

This is trickier though because type coercion from a string has a certain hierarchy (e.g. `float` is more stringent than `int`, `int` more stringent than `str`). The OBO file does not express this hierarchy, so `psims` would need to handle that, and it currently does not. This is doable, I'd just need an hour to figure out the ways it can go horribly wrong, test it, and make sure a fallback strategy is in place.

[...]

PR #165, opened 2025-03-14
Use psims to get cvParam type definitions

levitsky:
Addresses #161.

This divides `xml.XML` into classes, factoring out functionality related to handling of `UserParam` and `cvParam` elements. Parser dealing with `cvParam` now require `psims` and need to get or create a CV object. They then consult it when type information is not given in the attribute.

@mobiusklein could you please take a look and advise if this is going the right way? any edits or comments would be highly appreciated.

P.S. Upon posting I see that `MzML` behaves differently in `mzmlb` and I'm not sure why yet.


mobiusklein:
The approach seems reasonable. I had initially thought you'd directly use the value type conversion machinery in `psims`, but that would have made using the `unit*` types harder. I opened #166 to add a type cache to your implementation which should improve performance, given that the CV is essentially fixed at install time.

The issue with the mzMLb test is that when I wrote the mzMLb file, I was actually _reading_ the mzML file with `pyteomics` at the time, so the value types are invalid at the source there. I tried to use `msconvert` on `test.mzML` to re-convert it to mzMLb, but I got an unhelpful error: `[SpectrumList_mzML::create()] Bad istream`

levitsky:
To be honest, I just don't know enough about `psims` to think of any other way to use it, but if you say this makes sense, it's good. Should we think about entity inheritance in some way? I didn't see it directly taken care of in the code but I could easily have missed it. Is it possible that an entity has an `is_a` relationship to some other entity that has a `has_value_type`, for example?

My change makes parsers that just parse UserParams return regular types instead of `unit`-types, which is a subtle change but potentially breaking as well. Perhaps it would make more sense to make it depend on the element being a `cvParam` or a `userParam`, but I'm not really sure.

What would be the right way to take care of the `mzmlb` test? Can we perhaps use an externally-produced reference file for the test?

Is there anything else you would like changed in this PR?

Thanks again for your time.

mobiusklein:
`psims`'s type conversion logic is far from clean itself, rather hacked together to for `xsd` type coercion using a global lookup table. The way you're doing it here is probably better as it saves us from having `pyteomics` depend upon implementation details of `psims`. 

RE `userParams` type conversion, we can use the unit primitive wrappers without units, would that work? I can't tell quite where the conversion diverges right now.

RE mzMLb - https://github.com/mobiusklein/mzdata/blob/main/test/data/small.mzMLb was produced from https://github.com/mobiusklein/mzdata/blob/main/test/data/small.RAW by ProteoWizard, as was https://github.com/mobiusklein/mzdata/blob/main/test/data/small.mzML, so they should all agree.

I am a bit confused about why some unrelated tests are failing.

levitsky:
Thanks @mobiusklein, I updated the mzmlb test borrowing the small.mzMLb file from your repository (it's downloaded from within the test suite right now). What appears peculiar in the test data is that the resolving power term (at least its updated version) has a string type in the CV, but I guess the parser is working as intended, which is good. 

Regarding UserParam, right now each `ParamParser` and `CVParamParser` classes define their own `_param_types` and also have their own `_param_type` and related methods, however each of the classes handles `UserParam` and `cvParam` the same, `CVParamParser` producing unit types without actual unit info for `UserParam` elements.

The question is, what makes more sense for the user: have `UserParam` be handled slightly differently in different formats (current behavior) or have a mix of plain and unit types within the output of the same parser based on the type of element being parsed. The latter would arguably be more confusing to anyone who even notices the difference.

I am also not entirely happy with some code duplication between `_get_info` methods for `XML` and `ParamParser`

mobiusklein:
I still need to make time to review this again in more detail, but this detail might simplify things: In PSI formats, `userParam` can have CV-controlled units, even if the property itself isn't a CV parameter:

https://peptideatlas.org/tmp/mzML1.1.0.html#userParam
https://peptideatlas.org/PSI/schemas/mzIdentML/1.2/mzIdentML1.2.0.html#userParam
etc.

The OpenMS formats don't formally have units in their schemas.

levitsky:
Unit information should be extracted from mzML and mzIdentML `userparam`s with the current implementation, also I just added a small test for that.

What just occurred to me is that we still don't do any lookups for CV-controlled units in the ontology, nor do we check for `has_units` on the parameter terms. Now that we use the CV anyway, we could extract the unit name if only the unit accession is given, or retrieve unit information from `has_units` relationship without the units being specified in the file at all. Would that make sense?

mobiusklein:
Okay, that means that within a single format we shouldn't have something sometimes be a unit type and sometimes not? 

Give me until Sunday to review the code again in more detail please, I'm away at a conference this week. I'll see if there's a way to deal with the super-calls. 

RE unit resolution, we could check that a unit is valid for a given CV parameter, but outside of emitting a warning when an invalid unit is given or units are expected but missing, I don't see an immediate use-case for back-filling units. mzML does not stipulate _default_ units for anything. We broke with that in newer formats like mzSpecLib where the spec does specify default units so they can be omitted for brevity.

levitsky:
> Okay, that means that within a single format we shouldn't have something sometimes be a unit type and sometimes not?

This shouldn't happen. All userParams and cvParams should be either unit-aware or not, depending on parser class. There is also no check if something is a `userParam` or `cvParam`, both are processed in the same way. The only ontology the current `CVParamParser` implementation cares about is `PSI-MS`, so it could as well be called `PSIMSParamParser`. I don't know if this can or needs to be generalized, probably not at the moment.

> Give me until Sunday to review the code again in more detail please, I'm away at a conference this week. I'll see if there's a way to deal with the super-calls.

Absolutely, and thank you for your time.

> RE unit resolution, we could check that a unit is valid for a given CV parameter, but outside of emitting a warning when an invalid unit is given or units are expected but missing, I don't see an immediate use-case for back-filling units. mzML does not stipulate default units for anything. We broke with that in newer formats like mzSpecLib where the spec does specify default units so they can be omitted for brevity.

I would like to ask about these two specific things to make sure I understand.

Would you say it's impossible to have a cvParam in mzML or mzid, that has no unit information, but the corresponding term in PSIMS has a `has_units` relationship?

Also, the unit info extraction code in Pyteomics has a fallback to `unitAccession` when `unitName` is not present in the XML. Could we get the name from the CV in such cases, or do you think those cases should not happen at all and we don't need the fallback?

mobiusklein:
Apologies for the delay, the conference travel was interrupted with a family emergency.

> Would you say it's impossible to have a cvParam in mzML or mzid, that has no unit information, but the corresponding term in PSIMS has a has_units relationship?

It's not impossible, but the specification documents for those two formats provide no advice on how to proceed in such scenarios. In the trivial case, it is possible to back-fill units when the term has only one possible unit. For instance `collision energy` has only one unit `electronvolt`, so we could easily resolve this one. The same goes for many things that are m/z measurements like `selected ion m/z`. We are out of luck for any measurement of time, ion mobility, or intensity though, which can each pick from between four and six different units. Additionally, while a term might have a single unit today, it might have a choice of units in a year. 

Such files with missing information are syntactically valid, but probably would not be considered semantically valid.

> Also, the unit info extraction code in Pyteomics has a fallback to unitAccession when unitName is not present in the XML. Could we get the name from the CV in such cases, or do you think those cases should not happen at all and we don't need the fallback?

This is valid to do. We're not required to have both `unitName` and `unitAccession`, it's just convenient and nobody bothers to generate one but not the other. We would just look up `unitAccession` in the PSI-MS CV since it imports the Unit Ontology (or inlines it historically).

mobiusklein:
To the point about `_get_info` divergence, given how the overlapping concepts in `XML._get_info` and `ParamParser._get_info` diverge around the presence of the `params` list, we could unify the two methods by moving the conceptual blocks to helper methods that would be overridden by `ParamParser` but have `XML._get_info` simply pass around an empty `params` list.

It adds a little bit of overhead, and requires adding a few methods that are no-ops in `XML` (with their own overhead, again). If that's acceptable, that's a path forward to limit divergence in the future. It's not a total elimination of diverging logic, because we still need to add about five new "decision" points, but it's manageable.

levitsky:
Thank you and I hope your family emergency is/was resolved in the best way possible.

I can add a `unitAccesstion` lookup then as it seems easy to do, and if unit information is missing, we probably shouldn't do anything.

So if I understand correctly your point regarding the method inheritance we can basically move `_get_info` from `ParamParser` back to `XML` and implement all the helper methods there as no-op operations. I can see this working, I guess what stopped me from doing it this way was some ideological considerations along the line of - "if XML is not aware of parameter parsing logic, why would it have methods about it". I guess it is supposed to mean that the functionality is not properly encapsulated in this case. I don't know if it's a valid critique in any practical sense because I can see how this way is still better than completely independent `_get_info` implementations.

Perhaps using something similar to "element handlers" plug-in logic would be more optimal, with `ParamParser` defining special handlers for `*param` elements. But we already have a system of these, and it's probably not applicable here. Perhaps we could rewrite it slightly so that it can handle params as well as the current applications that it has?

levitsky:
After looking again at the inheritance / code duplication issue, it occurred to me that perhaps the easiest solution would be to not split `ParamParser` into a separate class at all and keep all this functionality in `XML`, as it was all this time. The new CV-awareness functionality is still in a subclass. This split was very easy to reverse.

Does it make sense like this @mobiusklein ?

mobiusklein:
Yes, that would accomplish what I was suggesting.

levitsky:
I'm going to merge this despite yet another hard-do-debug, not-locally-reproducible, unrelated and potentially temporary error in tests.

PR #171, opened 2025-04-28
Fix SQLAlchemy warnings

levitsky:
I took the liberty of trying to fix a few warnings SQLAlchemy has been throwing recently. I hope the fixes make sense but I am not experienced with SA and would appreciate a confirmation @mobiusklein


mobiusklein (approved):
This looks correct. I haven't used the newer SQLAlchemy version much though so my knowledge of the consequences will be limited.

PR #172, opened 2025-04-28
Make start method for map() configurable, avoid fork by default

levitsky:
Currently, the tests invoking `map()` method of readers produce `DeprecationWarning` on Linux:

```
/usr/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=59787) is multi-threaded, use of fork() may lead to deadlocks in the child.
  self.pid = os.fork()
```

That is because the implementation of `TaskMappingMixin` spawns both threads and processes, and the default method of spawning processes on Linux is currently `fork`, which is considered generally unsafe in a multi-threaded context. Currently `os.fork()` raises a warning when it detects multiple threads, and in Python 3.14 the default for Linux will apparently change to `spawn`, as it already is on Windows and macOS.

Discussion on the matter: https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555
See also: https://github.com/python/cpython/issues/84559

This change enforces the `spawn` context explicitly, which silences the warning and ensures identical behavior on all platforms and current and future Python versions. However, it will probably make things slower on Linux.

Any comments or alternative suggestions are welcome.

### DIFF
--- a/doc/source/data/indexing.rst.inc
+++ b/doc/source/data/indexing.rst.inc
@@ -164,3 +164,20 @@ and :py:meth:`iterfind`::
     controllerType=0 controllerNumber=1 scan=2
     controllerType=0 controllerNumber=1 scan=1
     controllerType=0 controllerNumber=1 scan=2
+
+Start method
+............
+
+There are different `methods to start new processes <https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods>`_,
+depending on the platform, and this will affect how the mapped function works (e.g. whether it can access global variables). Since version 5.0,
+Pyteomics allows you to configure the start method using :py:func:`pyteomics.auxiliary.set_start_method`.
+Unlike :py:func:`multiprocessing.set_start_method`, it will only set the start method for Pyteomics. If you instead want to call
+:py:func:`multiprocessing.set_start_method` and have it affect Pyteomics, you should make the call before importing Pyteomics.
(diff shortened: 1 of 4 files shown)


mobiusklein:
There is a potential for the mapped function to assume global variables are initialized by the parent process and continue to use them on the child process without copying them across explicitly. This will cease to work after `spawn` becomes the default. This has the potential to fail with a completely unhelpful error message too.

levitsky:
True, and I guess this is how `map()` has been working on Windows and Mac this whole time? 
Relying on globals being transparent to child processes has always been quite convenient on Linux, and for most of the time it works just fine; however, the core developers insist it is fundamentally unsafe to use `fork` like that.

I guess my question is, is `map()` going to be usable with `spawn`? The alternative seems to be to enforce `fork` in the future despite the warning.

Independently of the above, we can think about a threading version of `map()` to be used in free-threading Python.

mobiusklein:
No, `map` will still be usable, it just might not always provide the same amount of speedup it previously did on Linux. 

For free-threading Python, we can add a switch for controlling the parallelism backend. I'll need to read more about what its consequences are.

levitsky:
Correction: the new default for Linux will apparently be `forkserver` and not `spawn`: https://docs.python.org/3.14/whatsnew/3.14.html

Issue #176, opened 2025-06-20
Any way to get the composition from Proforma object?


levitsky:
### Discussed in https://github.com/levitsky/pyteomics/discussions/175

<div type='discussions-op-text'>

<sup>Originally posted by **retronade** June 20, 2025</sup>
I was going to calculate the isotopic distribution for all the precursors identified by DIA-NN. However, I can't work out how to obtain the composition from the Proforma peptide notation. While I can obtain the mass or m/z, Proforma is not a valid input for either mass.Composition or pyteomics.mass.mass.isotopologues. Consequently, I am struggling to utilise this helpful class/function. Is there a way to generate valid input for these from a Proforma object?

Thanks for writing this awesome library!</div>

PR #178, opened 2025-06-20
Implement ProForma.composition and fix adduct parsing

### LINKED ISSUE #177: ProForma parses adducts incorrectly
The current ProForma implementation does not correctly parse adduct information. I noticed this while looking into #176.

Example:

```python
In [1]: from pyteomics import proforma

In [2]: proforma.ProForma.parse('EMEVEESPEK/3[+2Na+,+H+]').charge_state
Out[2]: ChargeState(3, ['+2N+', '+H+'])  # note N+ instead of Na+
```

levitsky:
Fixes #177 and #176.

This fixes the adduct parsing problem while also making `ChargeState.adducts` a list of tuples for easier composition calculation. It also adds a `ProForma.composition()` which tries to calculate a neutral composition by default, but can also include adducts. This should be consistent with `Composition`, which is promised to stop supporting charge in the next release.

@mobiusklein any comments would be more than appreciated as it's my first time diving into the ProForma parsing machinery and I'm not very fluent in ProForma in general.


mobiusklein:
These bits are also changing in the upcoming version 2.1 of the specification, along with a lot of other changes as well. 98% of which will be backwards compatible. There are also a bunch of places where the specification wasn't precise enough and those places are being made formal but in different ways from how this parser handles them.

pyteomics/proforma.py
```diff
@@ -2534,3 +2563,56 @@ def find_tags_by_id(self, tag_id, include_position=True):
...
+        '''
+        comp = Composition()
+        if aa_comp is None:
+            aa_comp = std_aa_comp
+        try:
+            for aa, tags in self.sequence:
+                comp += aa_comp[aa]
+                for tag in tags or []:
+                    comp += tag.composition
+            for tag in chain(self.fixed_modifications, self.labile_modifications, self.unlocalized_modifications):
```
mobiusklein, line 2595: `fixed_modifications` should be checked when visiting each amino acid, similar to this:
https://github.com/levitsky/pyteomics/blob/f4033cb67f842102c6362a3b19fb14c811b6f801/pyteomics/proforma.py#L2337-L2340

pyteomics/proforma.py
```diff
@@ -2534,3 +2563,56 @@ def find_tags_by_id(self, tag_id, include_position=True):
...
+                comp += tag.composition
```
mobiusklein, line 2596: `tag.composition` may be `None` if you have a mass modification like `[+79.01]`. What to do in this case is kind of hard to answer, which is part of why I left it out this method. We could probably parameterize this like `error_on_missing_composition: bool = True` and let the user control whether this is an error or not.

mobiusklein (changes requested):

pyteomics/proforma.py
levitsky: Thanks, should be fixed now.

mobiusklein:
I don't know when v2.1 of the specification will be formally sent for review, beyond "soon", but I was planning to revise the parser and plug some holes that were formerly unspecified within the next month or two, time permitting.

I'll likely unspool from my current project by the end of the coming week and have time to work towards this. Another part of this will be to use this or a JS/WASM-target rewrite to build developer tools.

pyteomics/proforma.py
mobiusklein: Oh sure. I forgot that type didn't even stub out the attribute for duck typing.

All the CV-based modification tags implement `composition` as a wrapper around `self.definition.get("composition")` which would return `None` if the underlying dictionary doesn't have the key set, or if the key's value is `None`.

pyteomics/proforma.py
```diff
@@ -2592,7 +2592,10 @@ def composition(self, include_charge=False, aa_comp=None):
                 comp += aa_comp[aa]
                 for tag in tags or []:
                     comp += tag.composition
-            for tag in chain(self.fixed_modifications, self.labile_modifications, self.unlocalized_modifications):
+                for rule in self.fixed_modifications:
+                    if rule.is_valid(aa, self.n_term, self.c_term):
```
mobiusklein, line 2596: `rule.is_valid` has the signature `f(amino_acid: str, n_term: bool, c_term: bool)` where `n_term` and `c_term` indicate that the amino acid is at the N-terminal or C-terminal respectively. 

The for loop in `mass` uses the index to determine whether the current position is the N or C terminal:
https://github.com/levitsky/pyteomics/blob/3b2d87d00de3ea7a8002d6f2171b9748a32db28f/pyteomics/proforma.py#L2356-L2369

Specifically lines 2364 and 2365 tie these together.

levitsky: Oh, I am obviously not in the sharpest form working on this. Thanks for catching.

pyteomics/proforma.py
```diff
@@ -1579,22 +1578,22 @@ class ChargeState(object):
...
+            adducts = [('H+', charge)]
         self.charge = charge
         self.adducts = adducts
 
     def __str__(self):
         tokens = [str(self.charge)]
         if self.adducts:
             tokens.append("[")
-            tokens.append(','.join(str(adduct) for adduct in self.adducts))
+            tokens.append(','.join(f'{adduct[1]}{adduct[0]}' for adduct in self.adducts))
```
mobiusklein, line 1596: This may need to be adjusted to deal with the default case where the charge carriers are just protons and we just show a number. For instance, `PEPTIDE/2` and `PEPTIDE/2[+2H+]` are equivalent.

The adduct is one of the things being updated in v2.1 because the v2 specification is vague about how you write the adduct and how much charge an adduct contributes. See https://docs.google.com/document/d/1kD_BPpJluRJ5a6nqzDVvqMzs17NnMuHx/edit#heading=h.jiwt4vipxzug

levitsky: They do produce identical output currently:

```python

In [5]: str(proforma.ProForma.parse('PEPTIDE/2').charge_state)
Out[5]: '2[2H+]'

In [6]: str(proforma.ProForma.parse('PEPTIDE/2[+2H+]').charge_state)
Out[6]: '2[2H+]'
```

The changes in this part seem quite drastic, basically the adduct notation is completely different in v2.1.

levitsky: Ah, I see what you mean, it didn't occur to me that the `__str__` is used in `to_proforma`. This raises the question: should `to_proforma` be able to produce different version of ProForma, or should we just update it whenever v2.1 is adopted?

mobiusklein: We should probably focus on returning what the user provides by spec version, or at the level of the peptidoform.

mobiusklein (approved):

Issue #179, opened 2025-06-30
ProForma 2.1 Features


mobiusklein:
To implement v2.1, we need to do the following:

[] Charged formulas
[] Adduct and charge disambiguation
[] Implement modification localization rules
[] Better handling of unlocalized modifications
[] Named layers
[] Ambiguous colocalization control

We also should probably add a format version field to the `ProForma` object so we can keep track of which version(s) this is compatible with. Whether we try to robustly update to the latest format version is probably something we should control at the `to_proforma` function unless otherwise noted.

I think some of these can be handled incrementally, with adducts being the only major point of ambiguity, as everything else is either wholly backwards compatible or never occurs in earlier versions so we can't output something new from an existing input.

PR #183, opened 2025-07-29
ProForma parser refactor for ProForma 2.1

mobiusklein:
This rewrites the parser to use an object to manage data without piling dozens of variables in local scope and uses methods to isolate the handling of each state separately. It fixed a few bugs in the parsing of events pre-sequence without sigils (-, ^, ?), and adds support for:

- Charged formulae
- Sequence names (no strict ordering)
- Custom modification tags

It also parses but does not do anything with
- Ambiguous position limits (Limit, Position, CoMKP, CoMUP)
- Adducts

These components are still round-tripped back to string.

Realistically speaking, I don't see a way forward for those position limiters to enter the API without introducing a "peptidoform generator" algorithm which I refrained from doing previously, but am willing to go that direction

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -127,6 +127,10 @@ class CompositionNotFoundError(ProFormaError):
     pass
 
 
+class MissingChargeStateError(ProFormaError):
+    pass
+
+
 class UnknownMonosaccharideError(ProFormaError):
     pass
 
@@ -3530,7 +3534,7 @@ class ProForma(object):
         computing monoisotopic mass and uses the charge state to get the m/z.
 
         .. warning::
-            If no charge state of any kind is available, this will compute a neutral mass.
+            If no charge state of any kind is available, this will raise a :py:class:`MissingChargeStateError`.
 
         Parameters
         ----------
@@ -3554,7 +3558,7 @@ class ProForma(object):
                 f"Expected a charge state-like type, got {type(charge_state)}"
             )
         if charge_state is None:
-            raise ValueError(
+            raise MissingChargeStateError(
                 f"Requested an m/z value without providing a charge state and the peptidoform {self!r} does "
                 "not have a charge state itself."
             )


mobiusklein:
@levitsky this starts to address #179, but it is not complete. Nothing should be lost, but not all features are done.

levitsky:
Thank you. As far as I can see, this preserves ProForma 2.0 support while adding some 2.1 elements. Would you like me to try to add some tests for new features?

I'm not sure what you have in mind by "peptidoform generator" but I am happy to let you realize your idea.

mobiusklein:
By peptidoform generator, I mean a function/data structure that given a ProForma sequence with zero or more variably-specified modifications, generates every possible concrete peptidoform of that would satisfy those variable modification rules. Essentially like `parser.isoforms`, but inferring all the rules from ProForma as opposed to a separate instruction list.

I should write more tests myself, but if you have some usage ideas, I'd be happy to hear them.

As for test data, I'd likely go after the examples in https://github.com/douweschulte/proforma-grammar/blob/master/test.toml, but I don't know if my parser is quite robust enough for *all* of those. I don't have a good API in mind for things like the name layers and charges yet.

levitsky:
> By peptidoform generator, I mean a function/data structure that given a ProForma sequence with zero or more variably-specified modifications, generates every possible concrete peptidoform of that would satisfy those variable modification rules. Essentially like parser.isoforms, but inferring all the rules from ProForma as opposed to a separate instruction list.

That sounds extremely useful.

As for tests, I was just thinking about using the examples from the specification document. The test data collection is much richer though.

Name layers do seem tricky and "compound peptidoforms" even more so, if you are planning to even support them. The charges would now need to be localized on the formulas and then there should be a peptidoform property that sums all the localized charges and/or deals with the specified total charge, is that right?

Let me know whenever you want this merged so it doesn't slow you down.

mobiusklein:
All that was waiting was the adduct parsing, but I kept delaying because I was unhappy with the lack of a nice way to handle the incompatible syntax. Now it can read either 2.0 or 2.1, but only writes 2.1. I decided to take the compromise to get it done with.

levitsky:
Thank you! I was not aware that ProForma is now case-insensitive.

We could also switch this around and attempt to parse modX first. I think attempting to parse any non-trivial ProForma sequence as modX will fail, so we can support lowercase ProForma as well by attempting modX parsing first. The only problem is the edge case where a sequence is modX, but modifications are missing from `aa_mass`, leading to failure and then the sequence is interpreted as a valid (mixed-case) ProForma, leading to incorrect fragment series. Perhaps a warning for mixed-case ProForma can be emitted in places where we expect both modX and ProForma.

At this point I'm starting to think if maybe we need a standard sequence class, which is to say way to just instantiate a ProForma from a modX signature, and then we could use the same class everywhere internally. This would work best if we could encode arbitrary modifications by composition and without any database accessions. Would it work with a `FormulaModification`?

As for mass calculation, I think it should be able to account for adducts and charge as well. We have been historically conflating mass and m/z calculations for some reason. We could deduplicate this in 5.0 or keep it but I think we need a way to get m/z for ProForma using its encoded charge or user-provided charge.

Looking forward to hearing your thoughts on this, and as always, thanks a bunch.

mobiusklein:
At first I thought I could do a quick test for `[]{}<>?/` on the sequence to detect if there were ProForma sigils, but I realized that modX might define any of those as modifications

```python
>>> from pyteomics import parser
>>> parser.parse('[what if]^2[Spam]?[+105]-PEPT[+79.966331]IDE-[INFO:Oh no]')
['[what if]^2[Spam]?[+105]-',
 'P',
 'E',
 'P',
 'T',
 '[+79.966331]I',
 'D',
 'E',
 '-[INFO:Oh no]']

>>> parser.parse('[what if]^2[Spam]?[+105]-PEPT[Phospho]IDE-[INFO:Oh no]')
['[what if]^2[Spam]?[+105]-',
 'P',
 'E',
 'P',
 'T',
 '[P',
 'hospho]I',
 'D',
 'E',
 '-[INFO:Oh no]']
```

`parser.parse` will of course complain if you have unbalanced termini, but that's it. There are a few heuristics still, like counting if the # of [ == ] or equivalent, but they are all susceptible to errors. modX's power is that you're able to essentially define whatever notation you want as long as you adhere to the minimal syntactic rules. The drawback is you have to keep defining it everywhere you want to consume it.

ProForma was always case-insensitive, I just didn't bother to handle lower case amino acids because conventionally nobody uses them, but I was reading the spec as I wrote this so I included it as an afterthought. This is also why I think just special casing those internal API calls with case sensitivity re-enabled is reasonable.

As to parsing `modX` -> `ProForma`, yes a `FormulaModification` or `MassModification` would work, they are essentially there to be escape hatches for just this kind of scenario. The drawback with doing this is you either don't round-trip the `modX` sequence if you go directly into `ProForma`, or you subclass `ProForma` and make its `__str__` method serialize back to `modX`, which would require storing the original `modX` sigil in the modification tag's state, which isn't too bad, they have a loosely defined state dictionary for this reason. 

-------------------------

The `fragments` method has assumed that `charge` == 0 meant neutral mass, otherwise you specify the charge state you want. 

Adducts are complicated because some adducts are labile (e.g. ammonium, NH4+, HCD just breaks that down to an H+), and others are not (e.g. the metallic cations like Na, K, Ca[II]), where the convention has actually been to express those as modifications that can be localized. This was a difficulty I found when I worked with them in graduate school, and most proteomics tools lack a concept for adducts that aren't masquerading as modifications whose composition shift removes an number of hydrogens equal to the # of charges they confer. The only time I've seen this methodology fail is when dealing with elements whose nuclear binding energy falls outside the common organic range, like Cobalt. This is beyond my theoretical training so I don't have a good mental model for how to handle these.

Another complication is that when we talk about "neutral mass" of a peptide, we usually don't include any charge carrying protons in that quantity, so it is reasonable to assume that we don't want to include any other confirmed adducts. We do so implicitly when we compute an "m/z", but for most applications that's assumed to be `(neutral mass + charge carrier mass * charge) / charge`

levitsky (commented):
Thanks again for all this work. I took the liberty of trying to extend the composition fix a bit more, I hope it makes sense. Other than that I have just one small comment/question about the use of `ProformaError`.

Issue #189, opened 2026-01-27
ProteoformCombinator not working as expected


levitsky:
I wrote some tests for `proforma.ProteoformCombinator` and it doesn't seem to generate the output I expected. Do I not understand it correctly or is something amiss? @mobiusklein  I would appreciate you taking a look.

Note: the tests are now in `master` but to get to the actual behavior of `ProteoformCombinator`, we need to fix #187 first. #188 should fix it but I would appreciate a quick review.

PR #190, opened 2026-01-30
Support more varied forms of modification positioning in ProteoformCombinator

mobiusklein:
This adds extra behavior to to `ProteoformCombinator` to support modifications with uncertain positions defined by group IDs or for tagged intervals. It addresses the failing tests in #189.

I think this still needs work because the default behavior before assumed the user wanted both the modified and unmodified forms of a peptide, as you might get from a variable modification rule in a search engine. The failing tests assumed that was not the case. I think I can square these two use-cases more elegantly, but I only had about an hour to work on it tonight.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-5.0b4
+5.0b5
 -----
 
  - Support **ProForma 2.1** (`#183 <https://github.com/levitsky/pyteomics/pull/183>`_ by Joshua Klein).
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -4023,7 +4023,7 @@ class ProteoformCombinator:
                 self.template[i] = (aa, tags)
         self.template.fixed_modifications.clear()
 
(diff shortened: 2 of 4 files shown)


levitsky:
Thank you for looking into this.

I didn't think about the unmodified forms, but if they are produced, does it mean the full combinatorial variety of forms can be expected as well (like from a search engine)? In my tests I assumed that ProForma is written based on experimental evidence, such that e.g. the full m/z of the peptide is known, and so is the number of modifying groups. This interpretation is not based on reading anything so it can be completely invalid.

However in the current tests, the generator always produced one item only, which just left me completely puzzled as to what should be happening.

mobiusklein:
I had lost context on what I thought of as variable modification rules vs. what ProForma is used for. The current behavior of defaulting to excluding the case where a modification rule that could be applied is always applied is what is intended in ProForma.  The old behavior is still available, just not the default.

I also expanded the unit tests to include a more complicated use case where modification stacking was permitted.

PR #195, opened 2026-03-11
fix: PreIndexedMzML index type correction

### LINKED ISSUE #193: PreIndexedMzML incorrectly constructs inner indices
https://github.com/levitsky/pyteomics/blob/d90047926057ec615805ebe9590ad11741514893/pyteomics/mzml.py#L513-L515

This line automatically stores inner indices plain `dict` instead of `OffsetIndex`. I'll fix this one later this week.

mobiusklein:
Closes #193 

The inner index type should have been `OffsetIndex`, not a plain `dict`.


levitsky:
Thank you!

PR #196, opened 2026-03-12
feature: add proforma.modify_with

### LINKED ISSUE #194: Combinatorial (search-engine-like) proteoform generation in ProForma
cc @mobiusklein 

I have a use case for generation of all versions of a peptidoform with and without variable modifications applied, same way as `parser.peptidoforms` does for *modX*. We touched on this briefly in #190.  I was looking into the `ProForma` machinery to see if I can make use of the classes there. It appears that we would need to add a switch in `ProteoformCombinator.generate()` to do this, where instead of going over `itertools.product(*position_choices)` we would go over the product of all possible subsets of each item in `position_choices` item. Does it sound right?

Another question is the user-facing API for this. As far as I understand, the user can encode the rules into the sequence itself with something like `'[Oxidation|Position:M]?PEMPTIMDE'`, and that would work for `ProteoformCombinator` (assuming we implement the new iteration logic). However, no clean way seems to exist to pass external `variable_rules` to the constructor. So if I have a list of unmodified sequences and a set of modification settings, I need to either slap an `unlocalized_modifications` list onto each `ProForma` object, or manually create the rules for `ProteoformCombinator` like this:

```python
from pyteomics.proforma import (
    UnimodModification, ModificationRule, ModificationTarget,
    GeneratorModificationRuleDirective, ProteoformCombinator, ProForma
)

oxidation = UnimodModification('35')
rule = ModificationRule(oxidation, ModificationTarget('M'))
directive = GeneratorModificationRuleDirective(rule)
sequence = ProForma.parse('PEMPTIMDE')
combinator = ProteoformCombinator(sequence)
combinator.variable_rules = [directive]

for pf in combinator.generate(all_combinations=True):  # not implemented yet
   print(pf)
```

Is it the case that there is no friendlier way to set up `ProteoformCombinator` ? Do you think we can create some reasonable shortcut?

Potential ideas would be:
- document a way to instantiate a rule from a string, like `'[Oxidation|Position:M]'`, or just accept a mapping like `parser.peptidoforms`, creating the `ModificationTarget` and `ModificationRule` objects from keys and values;
- a way to pass a `list[GeneratorModificationRuleDirective]` to `ProteoformCombinator` constructor, or perhaps implicitly create them from `ModificationRule`s;
- a way to reuse a `ProteoformCombinator` instance for multiple sequences, or maybe an alternative class that only applies external modification rules/directives. Alternatively, we can just have a user-facing function encapsulating the code above. Then it doesn't really matter if it's slightly longer or shorter, although it would be nice to avoid any repeated parsing or cache its results.

I can try to put together an implementation but I wanted to hear your thoughts.

On a related note, currently if you specify `Position` for an unlocalized modification, `ProteoformCombinator` places this modification on possible sites and it still has the same `Position` modifier there. Is this desirable?

mobiusklein:
part of #194

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-5.0c0
+5.0c1
 -----
 
  - Update the standard ion compositions to be more consistent with the adopted ion type notation.
@@ -14,8 +14,10 @@
  - Support **ProForma 2.1** (`#183 <https://github.com/levitsky/pyteomics/pull/183>`_ by Joshua Klein).
    You can calculate compositions for :py:class:`ProForma` objects using :py:meth:`pyteomics.proforma.Proforma.composition`
    and get m/z with annotated or user-provided charge state using :py:meth:`pyteomics.proforma.Proforma.mz`.
-   You can also iterate through possible peptidoforms when a ProForma sequence is annotated with some ambiguity using
-   :py:meth:`pyteomics.proforma.Proforma.generate_proteoforms`.
+
+ - You can also iterate through possible peptidoforms when a ProForma sequence is annotated with some ambiguity using
+   :py:meth:`pyteomics.proforma.Proforma.proteoforms` and apply additional modification specifications to any ProForma sequence
+   using :py:func:`pyteomics.proforma.proteoforms` (`#196 <https://github.com/levitsky/pyteomics/pull/196>`_ by Joshua Klein).
 
  - Implement **thread-based parallelism**.
    Following the introduction of `official free-threading Python implementations <https://docs.python.org/3/howto/free-threading-python.html>`_
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -13,8 +13,8 @@ For more details, see the :mod:`pyteomics.proforma` online.
 import itertools
 import re
 import warnings
-from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, ClassVar, Sequence, Tuple, Type, Union, Generic, TypeVar, NamedTuple, DefaultDict
-from collections import Counter, deque, namedtuple
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, ClassVar, Sequence, Tuple, Type, Union, Generic, TypeVar, NamedTuple
+from collections import Counter, deque, namedtuple, defaultdict
 from functools import partial
 from itertools import chain
 from array import array as _array
@@ -4297,11 +4297,14 @@ class ProteoformCombinator:
     include_labile: bool
     variable_rules: List[GeneratorModificationRuleDirective]
 
-    def __init__(self, base_proteoform: ProForma, include_unmodified: bool=False, include_labile: bool=False, expand_rules: bool=False):
+    def __init__(self, base_proteoform: ProForma, include_unmodified: Optional[bool] = False, include_labile: bool = False, expand_rules: bool = False):
         if expand_rules:
-            if not include_unmodified:
+            if include_unmodified is False:
                 warnings.warn("Forcing `include_unmodified = True` from `expand_rules`")
             include_unmodified = True
+        else:
+            include_unmodified = bool(include_unmodified)
+
         self.template = base_proteoform.copy()
         self.include_unmodified = include_unmodified
         self.include_labile = include_labile
@@ -4381,12 +4384,12 @@ class ProteoformCombinator:
         return next(self._iter)
 
     def _invert_position_rules(self, rules: List[GeneratorModificationRuleDirective], positions: List[List[Optional[int]]]) -> List[List[Tuple[Optional[int], GeneratorModificationRuleDirective]]]:
-        index = DefaultDict(list)
+        index = defaultdict(list)
 
-        for rule, positions in zip(rules, positions):
+        for rule, position_list in zip(rules, positions):
             if rule.labile:
                 index[None].append(rule)
-            for position in positions:
+            for position in position_list:
                 if position is None:
                     continue
                 index[position].append(rule)
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -4160,6 +4160,11 @@ def peptidoforms(
         For all non-fixed modifications, include the case where the modification is not included anywhere.
     include_labile : :class:`bool`
         For all labile modifications, include the case where the modification is localized at every possible location.
+
+        .. note ::
+            Along with all possible localizations of the labile modifications, the output will contain peptidoform(s)
+            where it is kept as a labile modification without localization.
+
     expand_rules : :class:`bool`
         For all variable modifications, allow any number of copies of the modification to be included in the result.
         This mirrors the expected behavior of many search engines' variable modification rules, though it is not strictly
@@ -4310,6 +4315,11 @@ class ProteoformCombinator:
         in the input. See :attr:`expand_rules`.
     include_labile : :class:`bool`
         For all labile modifications, include the case where the modification is localized at every possible location.
+
+        .. note ::
+            Along with all possible localizations of the labile modifications, the output will contain peptidoform(s)
+            where it is kept as a labile modification without localization.
+
     strip : :class:`bool`
         If :class:`True`, the generated peptidoforms will have all modification tags stripped of any extra information,
         leaving only the bare modification definition.
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -1705,7 +1705,7 @@ class ModificationTarget(object):
                     t = tokens[0].lower()
                     n_term = t == "n-term"
                     c_term = t == "c-term"
-                    cls(tokens[1], n_term, c_term)
+                    return cls(tokens[1], n_term, c_term)
                 else:
                     raise PyteomicsError(
                         "Modification target has an invalid amino acid specific terminal target {1} in {0}".format(
@@ -1743,9 +1743,9 @@ class ModificationRule(object):
     modification_tag: TagBase
     targets: List[ModificationTarget]
 
-    def __init__(self, modification_tag: TagBase, targets: Union[ModificationTarget, List[ModificationTarget], None]=None):
+    def __init__(self, modification_tag: TagBase, targets: Union[ModificationTarget, List[ModificationTarget], List[str], None]=None):
         self.modification_tag = modification_tag
-        self.targets = targets
+        self.targets = targets  # type: ignore
         self._validate_targets()
 
     def is_not_specific(self) -> bool:
@@ -1764,10 +1764,11 @@ class ModificationRule(object):
         for target in self.targets:
             if isinstance(target, ModificationTarget):
                 validated_targets.append(target)
-            try:
-                validated_targets.append(ModificationTarget.from_str(target))
-            except PyteomicsError as err:
-                raise PyteomicsError(f"While parsing {self}, encountered error {err}") from err
+            else:
+                try:
+                    validated_targets.append(ModificationTarget.from_str(target))
+                except PyteomicsError as err:
+                    raise PyteomicsError(f"While parsing {self}, encountered error {err}") from err
 
         self.targets = validated_targets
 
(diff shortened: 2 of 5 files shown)


levitsky:
Thank you! I noticed a couple places that seemed to need fixing (I hope this makes sense, otherwise feel free to flag/roll back) and then I wanted to add some tests.
Before we get to implementing the full combinatorial proteoform generation, I thought we already had the ability to specify the number of occurrences of an unlocalized modification via `^`. While testing it, I noticed something strange: it seems to produce repeated peptidoforms.

```python
>>> for pf in proforma.ProForma.parse('[Oxidation|Position:M]^2?PEMPTIMDE').peptidoforms():
    ...:     print(pf)
    ...: 
PEM[Oxidation|Position:M]PTIM[Oxidation|Position:M]DE
PEM[Oxidation|Position:M]PTIM[Oxidation|Position:M]DE
```

This shows why a test I added is currently failing. I think the reason is simply that the Proforma parser supports the unclocalized modification count by adding multiple copies of the modification tag to the properties, and then the combinator treats them as distinct.

Do you see a way to fix this? This is not a problem with the new function, as you can see, but rather with the interplay of `ProteformCombinator` and the parser.

However, what I was initially trying to do was to see if we can support the count in the `peptidoroms` (`modify_with`) function in some way.

mobiusklein:
Well, that's a reminder to me to refresh my fork more often. Keeping proper track of repeated configurations was simple enough. The memory burden of tracking seen configurations should be small outside of some high complexity cases like trying to place dozens of different modifications on a full length protein sequence, but that's a bad case either way.

I think the tests are failing due to a network timeout unrelated to ProForma.

levitsky:
Thanks! This is indeed a simple solution (I was thinking along the lines of building a `Counter`-like object of variable mods and then handling the counts). For both solutions to work, though, it appears that we need to make `PositionModifierTag` hashable, which it is not yet.

So with this change the semantics of `include_unmodified` seems to expand to actually mean "include any subset of modifications", is that right? This means that combinatorial application is already all but implemented, we just need to supply the maximum count equal to the number of eligible sites.
 
```python
In [6]: seq = "[Phospho|Position:S|Position:T]^3?EMEVTSESPEK"

In [7]: pf = ProForma.parse(seq)

In [8]: for ppf in pf.proteoforms(include_unmodified=True):
   ...:     print(ppf)
   ...: 
EMEVTSESPEK
EMEVT[Phospho|Position:S|Position:T]SESPEK
EMEVTS[Phospho|Position:S|Position:T]ESPEK
EMEVTSES[Phospho|Position:S|Position:T]PEK
EMEVT[Phospho|Position:S|Position:T]S[Phospho|Position:S|Position:T]ESPEK
EMEVT[Phospho|Position:S|Position:T]SES[Phospho|Position:S|Position:T]PEK
EMEVTS[Phospho|Position:S|Position:T]ES[Phospho|Position:S|Position:T]PEK
EMEVT[Phospho|Position:S|Position:T]S[Phospho|Position:S|Position:T]ES[Phospho|Position:S|Position:T]PEK
```

This is great and perhaps `include_unmodified` should be renamed accordingly.

One thing that puzzles me is that this behavior does not seem to extend to `CoMUP` modifications:

```python
In [16]: seq = "[Phospho|Position:S|Position:T|comup|Limit:2]^2?EMEVTESPEK"
    ...: pf = ProForma.parse(seq)

In [17]: for ppf in pf.proteoforms(include_unmodified=False):
    ...:     print(ppf)
    ...: 
EMEVT[Phospho|Position:S|Position:T|CoMUP|Limit:2][Phospho|Position:S|Position:T|CoMUP|Limit:2]ESPEK
EMEVT[Phospho|Position:S|Position:T|CoMUP|Limit:2]ES[Phospho|Position:S|Position:T|CoMUP|Limit:2]PEK
EMEVTES[Phospho|Position:S|Position:T|CoMUP|Limit:2][Phospho|Position:S|Position:T|CoMUP|Limit:2]PEK

In [18]: for ppf in pf.proteoforms(include_unmodified=True):
    ...:     print(ppf)
    ...: 
EMEVTESPEK
EMEVT[Phospho|Position:S|Position:T|CoMUP|Limit:2]ESPEK
EMEVTES[Phospho|Position:S|Position:T|CoMUP|Limit:2]PEK
EMEVT[Phospho|Position:S|Position:T|CoMUP|Limit:2]ES[Phospho|Position:S|Position:T|CoMUP|Limit:2]PEK
```

In this case including the unmodified peptidoform also excludes the two colocalized peptidoforms.

mobiusklein:
Thank you for adding those extra test cases. I'll think about how I can improve the API for your first point about flexible upper bound on the number of times a rule is applied.

To your point about a `Counter`, yes, that absolutely is needed to make co-localization work. In the seven (or more?) other combinatorial modification expanders I've written, I explicitly refused to include colocalization because it makes the search space intractable. It has its uses though.

mobiusklein:
The refactor to use a `Counter` to track previous states makes `CoMUP` behavior work where it was lost.

I added a new `expand_rules` mode that can efficiently repeat modification rules on all positions, but it isn't compatible with co-localizating new variable modifications. There _should_ be a way to get to this working with `CoMUP`, but it's going to be complicated. If that is desirable, I can do it though.

levitsky:
Thank you! I am totally fine leaving the co-localization case for later as long as this limitation is documented.

I am still getting results contrary to my intuition playing around with the free `proteoforms` function converting rules from dict, for example in the `test_from_simple_dict` test case. With `include_unmodified=False`, shouldn't we get 3 forms (phospho on each of 3 sites), and with `include_unmodified=True`, an extra one for the unmodified form? Instead of 3 and 4, it produces just 2 and then 6.

So two questions about this are:

1. It looks like conversion from `{"Phospho": ['S', 'T']}` to `["Phospho|Position:S", "Phospho|Position:T"]` is not really equivalent, because we could get phospho on two S sites and we're not getting it.

2. If we are getting the non-modified, singly phosphorylated and some forms with 2 phospho with `include_unmodified`, then why not three as well? I'm having trouble distinguishing `include_unmodified` and `expand_rules` in this scenario.

Perhaps with the dict input we should just assume `expand_rules`? Or can we tweak the rules built from the dict input to make it produce some less surprising output?

P.S. The current assertions in the test don't have to stay this way, that was just me trying to reason about this.

mobiusklein:
I see your point. `{"Phospho": ["S", "T"]}` doesn't behave the same way that `["Phospho|Position:S|Position:T"]` does, even if I force it to, that's not the you are looking for. Additionally, if I move the logic for making it do so further into the guts of `ProteoformCombinator`, it won't be able to easily do its original job, which is to generate all the proteoforms defined by the ProForma sequence's own rules.

I'll implement what you suggested originally and see how it goes.

levitsky:
Moving `expand_rules` to the function only makes a lot of sense. I really like how the function works with string rule specifiers. I want to make sure in tests that it works correctly with multiple modifications and then I think we are good. What do you think?

Strange that the PROXI test failure is still not processed, I can try to look into that but it should not deter us.

mobiusklein:
I agree. When I tried to push that logic into the `ProteoformCombinator`, I was trying to make the system do something it wasn't meant to.

I'm still not completely happy with how `expand_rules` interacts with `CoMUP` and `Limit`, since it only copies the rule enough times to assign it to each position at most once, thus if you have two sites, `CoMUP|Limit:2` is applied, it can stack two modifications on one site and leave a second site unoccupied, not 0|0, 1|0, 0|1, 1|1, 0|2, 2|0, 2|1, 1|2, 2|2. But that is an excellent way to set off a complexity explosion. I think it'd be trivial to make it work (multiply copies by `rule.limit`), but I'm out of time before I need to get back to that Zstd PR and other work.

levitsky:
I was looking at it again trying going through the current tests to make sure I understand them, and I have a couple of notes/questions.

1. Not strictly related to this PR but I noticed that ProForma tests triggered some SQLAlchemy-related warnings that I remembered [fixing](https://github.com/levitsky/pyteomics/pull/171), and I realized `psims` ships its own version of `unimod` that doesn't have those changes. Would you be interested in updating that? It's not urgent at all because no errors are happening.

2. I was looking at `test_labile` and the semantics of `include_labile` and I was wondering the following. It currently generates all possible placements of the labile modifications and additionally one where it is kept as labile (in the simple case just copying the input proteoform). Wouldn't it make more sense to just produce all concrete peptidoforms? I feel like that's what the `proteoforms` method is supposed to do. It's great to have a switch for labile modifications but I have a feeling that the logic for labile modifications is currently a bit different than the unlocalized ones, for example.

As for covering colocalized modifications, your suggestion seems to have worked in my simple test. I don't know enough about ProForma to understand how the `max()` logic works with multiple limit specifiers, though.

mobiusklein:
I'll update `psims` to fix the deprecation warnings and get back to you about these other notes.

mobiusklein:
RE `labile` modifications. So labile modifications are a complicated topic that Proforma effectively says "This modification is present at intact precursor stage, but in the process of observing it, *we partially or completely destroy it* and it may not be localizable". This destruction is why a labile modification is both localized everywhere and also still included as not localized. 

I don't have a *good* way of expressing partial destruction. For the glycan composition case, it's theoretically viable to include every sub-composition at every position. For named modifications, it's basically ontology dependent. I could get neutral losses from Unimod, but I don't have a use case for this to test it with.

>As for covering colocalized modifications, your suggestion seems to have worked in my simple test. I don't know enough about ProForma to understand how the max() logic works with multiple limit specifiers, though.

That `max` expression is rather inscrutable, yes. `limit = max([t.value for t in tag.find_tag_type(TagTypeEnum.limit)] + [1])` says "find all the `Limit` tags on this modification and get their values as a list, and then concatenate that list with a list containing the default limit, `[1]`, so that if no `Limit` tag was specified, the default behavior is enforced and we don't have an empty list, then return the maximum value from the combined list with `max`".

levitsky:
Thank you for the explanation regarding labile modifications. The `max()` confusion was more about me not understanding what multiple `Limit` tags on a modification would mean.

I can look into slightly expanding documentation of `include_labile`, and add tests for more usage scenarios. Otherwise, the only thing I wanted to add is https://github.com/mobiusklein/pyteomics/pull/1, if it makes sense, or something to the same effect otherwise.

mobiusklein:
My apologies. Reading comprehension failure, followed by logic error.  The `max` is really arbitrary, and perhaps wrong. I wanted a succinct way to take either whatever `Limit` value the user provided, if any, or 1, and didn't want to explicitly check for the absence of an explicit limit. In ProForma, specifying multiple `Limit` tags on the same modification is unspecified, but the `max` selector means that the first or the second `Limit` will be chosen depending upon which is larger, not the order they were written. Since order is more likely to reflect intent, being inconsistent there is likely worse from a "law of least surprise" perspective.

The following is perhaps a more sane, less repetitive way to approach it:
```python
class TagBase:
    ...
    @property
    def limit(self) -> int:
        limit_tags = tag.find_tag_type(TagTypeEnum.limit)
        if limit_tags:
            if len(limit_tags) > 1:
                warnings.warn(f"{len(limit_tags)} Limit tags were found for {self}, this is undefined behavior. Only the first tag will be used")
            return limit_tags[0].value
        return 1
```
This preserves the _first_ `Limit` tag no matter the value.

The `strip` option makes sense to add, the targeting rules do make the sequence look more cluttered.

If you have suggestions about how to pose the labile modification issue, we're going to be discussing ProForma 2.1 at the PSI spring workshop in a week or two, and that would be a good time to bring it up.

levitsky:
Thanks again. At this point I would be happy to merge, I think we have a functional implementation.
It seems to be rather slow in practice so I would like to revisit it with some profiling but I am happy with the API functionality, and a release of Pyteomics is overdue by about a year now.
I have no thoughts on ProForma semantics or grammar with regards to labile modifications and I trust your understanding of it. 
Do you think ProForma discussions in the workshop can lead to changes in the format that would require changing the parser?

mobiusklein:
The discussion is unlikely to break existing syntax at this point. New
syntax might be added though.

The speed is an issue we can work on.

I'd say merge this and be done

PR #204, opened 2026-05-07
Optimize ProForma proteoform generation

levitsky:
This should preserve the functionality but improve speed by up to 50%.

One change is the addition of a `deepcopy` parameter that is disabled by default, assuming that the output of `ProteoformCombinator` will rarely be additionally modified.

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -3511,7 +3511,7 @@ class ProForma(object):
     def __len__(self):
         return len(self.sequence)
 
-    def __getitem__(self, i):
+    def __getitem__(self, i: Union[int, slice]):
         if isinstance(i, slice):
             props = self.properties.copy()
             ivs = []
@@ -3861,7 +3861,7 @@ class ProForma(object):
     def tags(self):
         return [tag for tags_at in [pos[1] for pos in self if pos[1]] for tag in tags_at]
 
-    def proteoforms(self, include_unmodified: bool = False, include_labile: bool = False, strip: bool = False) -> Iterator["ProForma"]:
+    def proteoforms(self, include_unmodified: bool = False, include_labile: bool = False, strip: bool = False, deepcopy: bool = False) -> Iterator["ProForma"]:
         """
         Generate combinatorial localizations of modifications defined on this ProForma sequence.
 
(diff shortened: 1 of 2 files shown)


pyteomics/proforma.py
```diff
@@ -4177,6 +4228,11 @@ def peptidoforms(
     strip : :class:`bool`
         If :class:`True`, the generated peptidoforms will have all modification tags stripped of any extra information,
         leaving only the bare modification definition.
+    deepcopy : :class:`bool`
+        If :class:`True`, the generated peptidoforms will be deep-copied from the original before modifications are applied.
```
mobiusklein: This probably means ``` :const:`True` ``` same with the false note below

levitsky: Sorry, I'm not entirely sure what you mean?

pyteomics/proforma.py
```diff
@@ -2961,6 +2982,22 @@ def parse(sequence: str, **kwargs) -> Tuple[List[Tuple[str, Optional[List[TagBas
         A mapping listing the labile modifications, fixed modifications, stable isotopes, unlocalized
         modifications, tagged intervals, and group IDs
     """
+    # short-circuiting the parser for simple sequences with no tags or modifications to avoid overhead
+    if sequence.isupper() and sequence.isalpha():
```
mobiusklein, line 3006: Good idea. It might make sense to wrap the empty properties in a `staticmethod` of `Parser`, but that might just be a succinctness issue

pyteomics/proforma.py
```diff
@@ -4019,6 +4060,37 @@ def create(self) -> TagBase:
...
+        for extra in tag.extra:
+            if extra.type == TagTypeEnum.position_modifier:
+                targets.append(extra.value)
+            elif extra.type == TagTypeEnum.comkp:
+                colocal_known = True
+            elif extra.type == TagTypeEnum.comup:
+                colocal_unknown = True
+        return targets, colocal_known, colocal_unknown
+
+    def _can_apply_with(self, tags) -> bool:
```
mobiusklein, line 4087: Good simplification

pyteomics/proforma.py
levitsky: Ah, the `:class:` label. Yes, of course, thank you

pyteomics/proforma.py
```diff
@@ -4467,7 +4510,7 @@ def generate(self):
                         state[((None, rule.token))] += 1
                         labile_remaining.append(rule.create())
                     continue
-                if idx not in rule.find_positions(template):
+                if not rule._can_apply_with(template.sequence[idx][1]):
```
mobiusklein, line 4533: I would bet this is the biggest source of speed up.

pyteomics/proforma.py
```diff
@@ -4413,49 +4478,27 @@ def __iter__(self):
...
         return position_choices
 
-    def _build_modification_iter(self) -> Iterator[List[Tuple[Optional[int], Optional[GeneratorModificationRuleDirective]]]]:
+    def _build_modification_iter(self) -> Iterator[Iterator[Tuple[Optional[int], GeneratorModificationRuleDirective]]]:
         position_choices = self._build_position_map()
         return map(lambda pos: zip(pos, self.variable_rules), itertools.product(*position_choices))
 
     def generate(self):
         seen = set()
         for slots in self._build_modification_iter():
```
mobiusklein, line 4516: When you have multiple modifications per position, `_build_modification_iter` produces a lot of redundant configurations. These are unlikely to be in the use-case you have in mind, but in the future if that becomes important, this is probably another place to tune.

mobiusklein (approved):
Thank you for reviewing this issue in detail. Your solutions look good, especially the place where I was being wasteful with nested `find_positions` calls.

pyteomics/proforma.py
levitsky: Yes, it's definitely up there, although I was benchmarking with only a Met oxidation, so there was not so many `slots` to begin with.

Issue #209, opened 2026-06-11
Support parsing chimeric ProForma (top-level `+`)


josiehong:
Hi @lev and @mobiusklein,

Thanks for all the work on pyteomics — the ProForma support in particular has been a huge help. I'd love to contribute chimeric parsing back, and wanted to check in on a couple of design choices before sending a PR.

ProForma 2.0 lets a chimeric spectrum be written as multiple peptidoforms joined by a top-level `+` (e.g. `EMEVEESPEK/2+ELVISLIVER/3`). Pyteomics recognizes the syntax but doesn't parse it — `proforma.parse("A+B")` raises `ProFormaError: ... Chimeric representation not supported`.

I have a working implementation, but since the parser already has reserved scaffolding for chimeric, I'd like your input on two design decisions before opening a PR:

```python
# ParserStateEnum
chimeric_start = 21
...
elif c == '+':
    raise ProFormaError(..., "Chimeric representation not supported")
```

## Decision A: where to split

- **In the state machine** — implement the reserved `chimeric_start` transitions (no duplicated depth-tracking; closest to the scaffolding's intent).
- **A standalone splitter** — split on top-level `+` while tracking `[] {} () <>` depth and the `/charge` edge case. Smaller, isolated diff, but re-derives logic the parser already has.

## Decision B: what a chimeric parse returns (backward-compat)

Independent of A:

- **New entry point only** — keep `parse()`/`ProForma` single-peptidoform; expose chimeric via e.g. `parse_any()` → `ProForma | ChimericProForma`. Non-breaking (downstream callers like depthcharge call `proforma.parse` directly and rely on its return type).
- **Transparent `parse()`** — `parse()` returns a container for `+` input. More discoverable, but changes the return type for existing callers.

Either way a chimeric result needs a container holding the component `ProForma`s — and a third, smaller question is what its name/shape should be (e.g. a `ChimericProForma` whose `mass`/`mz`/`fragments` return per-component tuples, vs. a dedicated result object).

Happy to open the PR once there's a direction on A and B. Thanks for the great library!

mobiusklein:
Hello,

I thought I recognized your name when you spoke at ASMS. That was a good description of the problem of de novo sequencing for two or more precursors.

[...]

PR #210, opened 2026-06-12
Add opt-in support for chimeric ProForma parsing

josiehong:
Adds opt-in parsing support for chimeric ProForma strings using `chimeric=True` related to #209. 

This keeps the default single-sequence behavior unchanged while allowing top-level `+` to separate chimeric peptidoform components when explicitly requested.

## Changes

- Add `chimeric` option to `Parser`, `proforma.parse`, and `ProForma.parse`.
- Return a list of parse results / `ProForma` objects when `chimeric=True`.
- Share fixed modifications, isotope labels, and peptidoform names across chimeric components.
- Preserve the current default behavior for `+`, with an error message suggesting `chimeric=True`.
- Add tests for opt-in parsing, shared metadata, empty components, charge/adduct boundaries, and shared fixed modifications.
- Document chimeric parsing and component-wise use with existing downstream APIs.

### DIFF
--- a/tests/test_proforma.py
+++ b/tests/test_proforma.py
@@ -138,6 +138,60 @@ class ProFormaTest(unittest.TestCase):
             self.assertEqual(i.charge_state.charge, charge)
             self.assertEqual(i.charge_state.adducts, adducts)
 
+    def test_chimeric_parse_requires_opt_in(self):
+        with self.assertRaisesRegex(ProFormaError, 'chimeric=True'):
+            parse('PEPTIDE+ELVIS')
+
+    def test_chimeric_parse(self):
+        seq = 'EMEVEESPEK/2+ELVISLIVER/3'
+        parsed = parse(seq, chimeric=True)
+        self.assertEqual(len(parsed), 2)
+        self.assertEqual(parsed[0][1]['charge_state'].charge, 2)
+        self.assertEqual(parsed[1][1]['charge_state'].charge, 3)
+
+        forms = ProForma.parse(seq, chimeric=True)
+        self.assertEqual(len(forms), 2)
+        self.assertTrue(all(isinstance(form, ProForma) for form in forms))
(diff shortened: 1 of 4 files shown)


mobiusklein:
@levitsky can you please approve workflow runs for this PR? I don't have permission.

tests/test_proforma.py
```diff
@@ -138,6 +138,60 @@ def test_charge_adducts(self):
...
+        self.assertEqual(len(parsed[1][1]['fixed_modifications']), 1)
+
+        forms = ProForma.parse(seq, chimeric=True)
+        aa_comp = std_aa_comp.copy()
+        aa_comp['cam'] = Composition(formula='H3C2NO')
+        self.assertEqual(forms[0].composition(), Composition(sequence='AcamC', aa_comp=aa_comp))
+        self.assertEqual(forms[1].composition(), Composition(sequence='camCcamC', aa_comp=aa_comp))
+
+    def test_chimeric_shared_names_and_isotopes(self):
+        parsed = parse('(>sample)<13C>AC+CC', chimeric=True)
```
mobiusklein, line 183: This bit is tricky. Names defined with `>` and `>>` are specific to a single peptidoform and peptidoform ion respectively. Only `>>>`  names apply to all peptidoforms in the set. This is section 8.2 of the forthcoming specification: https://docs.google.com/document/d/1HaSOMNmdFCxy0G6lq1Yci03z-4esmQoe_hSyoqTeRnc/edit?tab=t.0

So `AC` would have the name `sample`, but `CC` would have no name.

The example from the specification is `(>Trypsin)AANSIPYQVSLNS+(>Keratin)AKEQFERQTA`

mobiusklein (approved):
This looks great other than the name sharing thing (which is trivial and can be fixed at a later date).

I'm happy to bundle that fix in with other changes when I do a coverage overhaul.

PR #212, opened 2026-06-29
Fix numerous logic errors from historical lack of tests

### LINKED ISSUE #211: Support for unresolved modification tags in ProForma
@RalfG were there other breaking changes in v5?

_Originally posted by @mobiusklein in https://github.com/levitsky/pyteomics/issues/208#issuecomment-4636981951_
            
Hi @mobiusklein,

The main issue we've found so far is the following: In psm_utils, we used `pyteomics.proforma`'s resolving functionality to map arbitrary modification labels to resolvable modifications. In Pyteomics v5, this is not possible anymore, as only fully ProForma-legal labels are now accepted. Also see compomics/psm_utils#151. 

Not sure what the best way forward is, but I'm flexible in changing things in psm_utils if that makes sense.

Thanks!
Ralf

mobiusklein:
Addresses the lingering issue from #209

Fix #211, making tag resolution failure a non-fatal error at parse time. Still invokes resolvers, but doesn't crash if they fail.

Amongst other things, I plugged a bunch of holes in the hand-rolled state machine where state transitions didn't exist, or had muddled invariant management for bizarre scenarios.

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -3251,6 +3251,9 @@ class Parser:
                 self.index,
                 self.state,
             )
+        for v in self.intervals:
+            assert v.start is not None
+            assert v.end is not None
         return self.positions, {
             "n_term": self.n_term,
             "c_term": self.c_term,
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -67,7 +67,7 @@ class Chimeric(Generic[T], Sequence[T]):
 
     __slots__ = ('peptides', 'chimeric')
 
-    __match_args__ = ["peptides", "chimeric"]
+    __match_args__ = ("peptides", "chimeric")
 
     def __init__(self, peptides: List[T], chimeric: Optional[bool]=None):
         self.peptides = peptides
@@ -3374,7 +3374,7 @@ def parse(
             Parser.empty_properties()
         )
         if chimeric:
-            return [result]
+            return Chimeric([result], chimeric=False)
         return result
     parser = Parser(sequence, chimeric=chimeric, **kwargs)
     return parser.parse()
@@ -4081,6 +4081,7 @@ class ProForma(object):
         """
         result = parse(string, chimeric=chimeric, **kwargs)
         if chimeric:
+
             return Chimeric([cls(*component) for component in result], result.chimeric)
         return cls(*result)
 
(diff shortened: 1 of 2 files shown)


levitsky:
Thank you! I see that you also implemented a `Chimeric` type.

The tests pass for me locally on Python 3.12 but they do fail on 3.11:

```python
$ python -m unittest test_proforma.ProFormaTest.test_fragments
E
======================================================================
ERROR: test_fragments (test_proforma.ProFormaTest.test_fragments)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/levitsky/pyteomics/tests/test_proforma.py", line 166, in test_fragments
    for o, e in zip(i.fragments("b"), b_expected):
                    ^^^^^^^^^^^^^^^^
  File "/home/levitsky/pyteomics/pyteomics/proforma.py", line 4222, in fragments
    intervals = sorted(intervals, key=lambda x: x.start, reverse=reverse)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object cannot be interpreted as an integer

----------------------------------------------------------------------
Ran 1 test in 0.002s

FAILED (errors=1)
```

mobiusklein:
Okay, no clue where that `None` to `int` coercion is happening. Either cosmic rays or non-deterministic behavior. All set on my end though.

levitsky:
Thank you. Before merging, I want to try it out and add some more tests potentially around `Chimeric`. For example, this doesn't work as I expected:

```python
In [1]: from pyteomics import proforma

In [2]: proforma.ProForma.parse('PEPTIDE', chimeric=True)
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 proforma.ProForma.parse('PEPTIDE', chimeric=True)

File ~/pyteomics/pyteomics/proforma.py:4084, in ProForma.parse(cls, string, chimeric, **kwargs)
   4082 result = parse(string, chimeric=chimeric, **kwargs)
   4083 if chimeric:
-> 4084     return Chimeric([cls(*component) for component in result], result.chimeric)
   4085 return cls(*result)

AttributeError: 'list' object has no attribute 'chimeric'
```
I will have more time next week to look into it.

mobiusklein:
I missed the fast path, thank you for checking that. I also realized I could actually use pattern matching freely in unit tests now that Py310 is the minimal version, which caught another syntactic issue.

levitsky (approved):
Thank you!

I noticed an issue with `ProteoformCombinator` that I missed before because I didn't bother to test fixed modifications. I took a guess at a fix because the semantics of those `ModificationRuleDirective` creation methods and the class itself are not entirely clear to me, and multiple errors shown by static type checkers don't help.

Could you check if `ProteoformCombinator._apply_fixed_modifications` makes sense?

levitsky:
Thanks again! Should I merge now?

mobiusklein:
Yes please.

PR #213, opened 2026-07-12
fix: correct tuple index in ProForma.__getitem__ slicing

steps-re:
## Bug

Slicing a `ProForma` sequence that has a grouped modification (e.g. `#g1`) raises a `TypeError`.

In `ProForma.__getitem__`, when handling a slice, the code loops over `tag_hits` from `find_tags_by_id(group_id, include_position=True)`, which returns a list of `(position, tag)` tuples. The loop variable `i` gets reassigned to `tag_hits[0]`, the whole tuple, and is then used directly as `subseq[i]`. `__getitem__` only accepts `int` or `slice`, so indexing with a tuple fails.

Repro:

```python
from pyteomics import proforma

seq = "EMEVT[#g1]S[#g1]ES[#g1]PEK"
obj = proforma.ProForma.parse(seq)
obj[2:9]
# TypeError: list indices must be integers or slices, not tuple
```

## Fix

Use `tag_hits[0][0]` (the position) instead of the full tuple.

## Testing

- Added `test_slice_grouped_modification` in `tests/test_proforma.py`, which fails with the `TypeError` above on the old code and passes with the fix.
- Ran the full `tests/test_proforma.py` suite: 62 passed, 0 failed, no regressions.


mobiusklein:
Thank you for reporting this. I believe this is amongst the many historical bugs being fixed in #212. Manually testing your example with that branch worked as expected. Could you please test that branch for your use-case?

steps-re:
Confirmed, thanks @mobiusklein. I checked out `fix/proforma-coverage-fixes` (#212) and ran my exact repro against it:

```python
from pyteomics import proforma
sub = proforma.ProForma.parse("EMEVT[#g1]S[#g1]ES[#g1]PEK")[2:9]
# -> "EVT[#g1]S[#g1]ES[#g1]P", group_ids == ['#g1']
```

No more `TypeError` and the slice is correct, so #212 covers this. Closing this in favor of it.

If it's useful, the regression test I added here (`test_slice_grouped_modification`) pins this specific grouped-modification slice case, and #212 is all about backfilling tests for historically untested paths, so you're welcome to lift it in. Happy to open a tiny test-only PR against your branch if you'd prefer. Thanks for the fast turnaround.

PR #214, opened 2026-08-01
RESID in ProForma

acesnik:
Hi there, I've been working on a tool that takes in ProForma notation, and I encountered some issues using RESID here. I drafted up a PR with the help of an LLM, depending on a request to add a new `load_resid()` method to `psims` to correct RESID loading there (https://github.com/mobiusklein/psims/pull/26) that p. Thanks for considering it!

### DIFF
--- a/doc/source/api/proforma.rst
+++ b/doc/source/api/proforma.rst
@@ -83,7 +83,7 @@ coupled with minimal information about mass and position data.
 Dependencies
 ------------
 
-To resolve PSI-MOD, XL-MOD, and GNO identifiers, :mod:`psims` is required. By default,
+To resolve PSI-MOD, XL-MOD, GNO and RESID identifiers, :mod:`psims` is required. By default,
 :mod:`psims` retrieves the most recent version of each controlled vocabulary from the internet, but
 includes a fall-back version to use when the network is unavailable. It can also create
 an application cache on disk.
@@ -134,7 +134,7 @@ These features are independent from each other:
 
 3. Top Down Extensions
 
-    - [ ] Additional CV/ontologies for protein modifications: RESID (the prefix R MUST be used for RESID CV/ontology term names)
+    - [x] Additional CV/ontologies for protein modifications: RESID (the prefix R MUST be used for RESID CV/ontology term names)
     - [x] Chemical formulas (this feature occurs in two places in this list).
 
 4. Cross-Linking Extensions
@@ -198,6 +198,8 @@ Modification Tags
 
 .. autoclass:: PSIModModification
 
+.. autoclass:: ResidModification
+
 .. autoclass:: XLMODModification
 
 .. autoclass:: GNOmeModification
@@ -240,5 +242,7 @@ Modification Resolvers
 .. autoclass:: PSIModResolver
 
 .. autoclass:: XLMODResolver
+
+.. autoclass:: ResidResolver
 
 .. autoclass:: GNOResolver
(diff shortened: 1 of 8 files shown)


mobiusklein:
The relevant `psims` change has been released, so this should be ready to proceed.

pyteomics/proforma.py
```diff
@@ -872,6 +875,58 @@ def _resolve_impl(self, name=None, id=None, **kwargs):
         return rec
 
 
+class ResidResolver(ModificationResolver):
```
mobiusklein, line 878: Do you want this to be wired into the `GenericModification` resolution stack as well, so you can just write `[AA0038]` and have the name resolve correctly? If so, add it to the `CustomModification`'s default resolver.

acesnik: 🤖

Done — `ResidModification.resolver` now sits in `GenericModification`'s chain, after `GNOmeModification` and before the non-strict Unimod pass. That ordering keeps it additive; across all 619 RESID entries:

```
bare accessions   0 resolved before, 619 now
names             587 unchanged, 32 newly resolvable, 0 changed
```

So `[AA0038]` resolves, and `[O-phospho-L-serine]` still comes from PSI-MOD. Tests both ways.

I read your suggestion as the generic chain rather than `CustomModification`'s, since `CustomResolver` is the register-your-own store — easy to move if you meant that one.

acesnik:
🤖

psims 1.4.0 is out, so `test-requirements.txt` now asks for it and CI exercises RESID instead of skipping it. Out of draft.

doc/source/api/proforma.rst
```diff
@@ -134,7 +134,7 @@ These features are independent from each other:
 
 3. Top Down Extensions
 
-    - [ ] Additional CV/ontologies for protein modifications: RESID (the prefix R MUST be used for RESID CV/ontology term names)
+    - [x] Additional CV/ontologies for protein modifications: RESID (the prefix R MUST be used for RESID CV/ontology term names)
```
levitsky, line 137: If I'm reading the conversation right, the part about mandatory prefixes can be updated?

acesnik: Hmm, I'd leave it as it stands. The checklist is quoted from the spec, and prefix rules are about what an implementation writes. R: still round-trips as RESID:. The XL-MOD and GNOme lines carry the same wording and already resolve unprefixed names, too.

pyteomics/proforma.py
levitsky: Makes sense to me. @mobiusklein is this what you had in mind?

mobiusklein: Yes, this is correct. This is also out of spec but is consistent with the user's expectations.

mobiusklein: Working through the chain of events, this _might_ actually break things where it wouldn't before. The scenario is when a user has `psims` installed, but not a new enough version that they have RESID support. Their existing code uses `pyteomics.ProForma`, and can load PSI-MOD and Unimod through it. This lets them use `GenericModification` previously. After they upgrade `pyteomics`, but not `psims`, they will attempt to parse a sequence with a generic modification, and the resolver chain will hit RESID and crash with an exception before it reaches the end of the list where the relaxed Unimod resolver is.

So we have two, maybe three options:
1. Set a minimum `psims` version on the optional dependency, not the test requirement
2. Remove `ResidResolver` from the generic chain
3. Move `ResidResolver` to the end of the list. 

I think  option 1 is the sanest thing to do.

levitsky: I agree 1 is the sanest option. As a potential option 4, we could probably let the chain of trials in `GenericResolver` gracefully handle the exception raised by `_needs_newer_psims`

levitsky: Updated the `psims` version requirement

levitsky:
Thank you @acesnik and @mobiusklein!

PR #215, opened 2026-08-02
ProForma support for inter-chain cross-linked peptides

acesnik:
Hi there — another PR for inter-chain crosslinks, to round out the last unchecked box in the ProForma implementation. A peptidoform written with `//` gives you a `ProForma` whose new `chains` property holds more than one entry. There's no new type for a cross-linked peptide, and `parse()` and `ProForma.parse()` return exactly what they did before, with no new parameters. (I used an LLM to assist with the code and the description below.) Thanks for considering it!

---

The last unchecked box on the ProForma compliance list reads "cross-linked peptides (using the XL-MOD CV/ontology)", which undersells what is already here: XL-MOD resolves, the X prefix works, and cross-links within one peptide compute the right mass. What was missing is the // notation of section 9.2.2, joining two or more chains into one covalently bonded molecule, refused outright by both parsers.

ProForma is a three-level hierarchy -- peptidoform ion set (+), peptidoform ion (//), peptidoform -- and this modeled the outer and inner levels with Chimeric and ProForma, with nothing between. A string using // yields a ProForma whose `chains` holds more than one entry. What parse() and ProForma.parse() return is unchanged: ProForma, or Chimeric[ProForma]. 

`chains` is the accessor. Quantities belonging to the whole molecule span it -- mass, charge state, and str(), which rejoins with // and writes the (>>name) and (>>>name) prefixes highest-first as the specification requires. Operations with no meaningful multi-chain answer refuse rather than silently answering for the first chain: fragments() and proteoforms() raise, naming `.chains`, as does find_tags_by_id() when positions are requested. composition() and mz() span the chains, since the arithmetic is the same as mass.

A cross-link is one molecule however many sites it bridges, so its mass is counted once however many sites declare it. Where the linker is written once and the other sites carry a bare [#XL1] that falls out of the arithmetic, since a bare reference has no mass. Two cases needed handling:

  - Across chains, repeating the declaration is useful: it lets a chain be read on its own without hunting the others for what #XL1 refers to, and section 9.2.2 writes its examples both ways. Both denote one molecule, so they now agree rather than differing by a linker mass. This reads the blanket requirement in 9.2 that a cross-linker be "mentioned once only" as being about self-links in 9.2.1.

  - Within one peptide a repeat is redundant, which is what 9.2.1 is about. ProForma.mass discounts it too and warns, so the two agree. The scan is guarded on a #XL group being present, leaving the common path as it was.

ProFormaParseResult stops being an alias for a bare tuple and becomes the type that carries an ion's chains. For a single chain it still unpacks and compares as the (positions, properties) pair it was, so `positions, properties = parse(s)` is unaffected.

Eleven previously commented-out cases now run, covering inter-chain cross-links, #BRANCH, disulfides, chimeric sets of cross-linked ions, and the full Trastuzumab example with its five inter-chain disulfides and all three naming levels.

Two fixes fell out of the work. TagTypeEnum.xlmod was missing from TagBase.is_modification, so find_modification could not see a cross-linker at all. And the commented-out insulin case was a truncated copy of the specification's example: everything before the A chain's last residue had been lost, taking the #XL1 and #XL2 definitions with it and leaving two references to groups that were never defined. Restored from section 9.2.3.

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -136,6 +136,43 @@ def _cross_link_overcount_within(chain) -> Tuple[float, List[str]]:
     return excess, sorted(duplicated)
 
 
+def _cross_link_declared_compositions(chain) -> Dict[str, List[Any]]:
+    """Compositions declared for each cross-link group within one chain."""
+    declarations: Dict[str, List[Any]] = {}
+    for _position, tags in getattr(chain, 'sequence', ()) or ():
+        for tag in tags or ():
+            group = getattr(tag, 'group_id', None)
+            if not (group and str(group).startswith('#XL')):
+                continue
+            if not (tag.is_modification() and tag.has_composition()):
+                continue
+            try:
+                composition = tag.composition
+            except Exception:
+                continue
+            if composition is not None:
+                declarations.setdefault(str(group), []).append(composition)
+    return declarations
+
+
+def _cross_link_composition_overcount(chains):
+    """The composition counterpart of :func:`_cross_link_overcount`.
+
+    A linker declared on several chains is one molecule, so its atoms are
+    counted once, exactly as its mass is.
+    """
+    seen = set()
+    excess = Composition()
+    for chain in chains:
+        within = _cross_link_declared_compositions(chain)
+        for group, compositions in within.items():
+            start = 0 if group in seen else 1
+            seen.add(group)
+            for composition in compositions[start:]:
+                excess += composition
+    return excess
+
+
 def _cross_link_overcount(chains) -> Tuple[float, List[str]]:
     """Cross-linker mass counted more than once, and the groups responsible.
 
@@ -2876,13 +2913,13 @@ class Parser:
 
     def _close_ion(self):
         """End the peptidoform ion under construction."""
-        self.chains.append(self._finish_component())
+        self.chains.append(self._finish_chain())
         self.components.append(self.chains)
         self.chains = []
 
     def _handle_chain_separator(self):
         self.saw_chain_separator = True
-        self.chains.append(self._finish_component())
+        self.chains.append(self._finish_chain())
         self._reset_component()
 
     def _handle_chimeric_separator(self):
@@ -3346,7 +3383,7 @@ class Parser:
             self.index += 1
         return self.index < self.length
 
-    def _finish_component(self) -> PeptidoformIon:
+    def _finish_chain(self) -> ProFormaParseResult:
         if self.charge_buffer:
             charge_number = self.charge_buffer()
             if self.adduct_buffer:
@@ -4183,7 +4220,9 @@ class ProForma(object):
         elif other is None:
             return False
         else:
-            return self.sequence == other.sequence and self.properties == other.properties
+            return (self.sequence == other.sequence
+                    and self.properties == other.properties
+                    and self.additional_chains == getattr(other, 'additional_chains', []))
 
     def __ne__(self, other):
         return not self == other
@@ -4403,7 +4442,6 @@ class ProForma(object):
         -------
         float
         """
-        self._single_chain("mz()")
         charge_state = charge
         if charge_state is None:
             charge_state = self.charge_state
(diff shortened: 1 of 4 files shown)


pyteomics/proforma.py
```diff
@@ -4163,6 +4403,7 @@ def mz(self, charge: Union[int, ChargeState, None] = None, **kwargs) -> float:
         -------
         float
         """
+        self._single_chain("mz()")
```
levitsky: This has me a bit confused. The mass and charge properties seem to be supported, but not m/z?

pyteomics/proforma.py
```diff
@@ -4446,6 +4689,7 @@ def composition(self, include_charge: Union[bool, ChargeState]=False, aa_comp=No
         Composition
             :py:class:`Composition` object representing the composition of the ProForma sequence.
         '''
+        self._single_chain("composition()")
```
levitsky: Is there a reason this cannot work?

pyteomics/proforma.py
```diff
@@ -3214,7 +3346,7 @@ def step(self) -> bool:
             self.index += 1
         return self.index < self.length
 
-    def _finish_component(self) -> ProFormaParseResult:
+    def _finish_component(self) -> PeptidoformIon:
```
levitsky: This method does not return a `PeptidoformIon`. As far as I understand, it actually returns a single chain, while `components` refers to independent (possibly multi-chain) parts of a chimeric annotation. The value of this method is appended to `chains`, so I guess the name of the method should be changed to `_finish_chain` or similar and the return type must be `ProFormaParseResult` (which we might need to consider renaming as well).

levitsky (commented):
Thank you for this contribution. With review, I would defer to @mobiusklein as someone who actually knows and understands the specification and the parser; that said, I did my best to look it through and left a couple of comments.

mobiusklein:
This is going to take me a little bit to get to as the cross-linking part of ProForma is not something I paid careful attention to since I have never worked with crosslinkers myself.

I am working on understanding why a crosslinked peptide needs its own separate type, rather than keying off a new property that points to attached chains. The methods themselves would certainly need to check and handle those cases differently.

pyteomics/proforma.py
acesnik: 🤖

Fixed. `mz()` only refused because `composition()` did.

pyteomics/proforma.py
acesnik: 🤖

You're right that it can. It spans the chains now — the same arithmetic as `mass`: chains summed, each cross-linker counted once however many chains declare it. Composition and `.mass` agree exactly.

pyteomics/proforma.py
acesnik: 🤖

Renamed `_finish_chain`, as you suggested.

One divergence worth flagging: you said the return type must be `ProFormaParseResult`, and it's `ProFormaChain`. I took your parenthetical and did the rename, which fixes both things at once — the per-chain pair gets a name that says *chain*, and `ProFormaParseResult` is freed to mean what `parse` returns, which is the name that had been misread as an ion-tier type. Happy to put it back.

acesnik:
🤖

@mobiusklein — there's no separate type; my description was misleading and I've fixed it. `ProForma` gained a `chains` property, and a `//` string gives you a `ProForma` with more than one entry in it, with `parse()` returning what it did before. That's the property-pointing-at-chains shape you described. An earlier draft did put a container above `ProForma`, and it pushed the hierarchy onto every caller, so it went.

The class left in the diff is only the low-level `parse()` result. It was `PeptidoformIon`, which read as an ion-tier container and so implied the design I'd dropped — that's what tripped you, and @levitsky separately. Now `ProFormaParseResult`.

Three more bugs turned up from generalising @levitsky's comments, all silent: `copy()` dropped every chain but the first, a two-chain ion compared equal to its own first chain, and `find_tags_by_id` searched only the first chain — which matters most for cross-links, since the two ends sit on different chains. Fixed, with tests.

levitsky:
I did my best to reconcile this branch with the updated master and fix issues after two other ProForma-related PRs were merged. @mobiusklein could you take a look at it again?

PR #217, opened 2026-08-13
Enforce terminal grammar boundaries

BioGeek:
## Summary

Tighten the ProForma parser's handling of terminal modification groups and completed constructs so that it follows the HUPO-PSI grammar rather than accepting and silently discarding invalid input.

The change adds a deterministic grammar-conformance matrix that combines valid optional prefix, sequence, and suffix productions, plus a matrix of invalid grammar mutations.

### Fixed behaviour

- Reject a second N-terminal modification group, rather than overwriting the first group's modifications. (Supersedes https://github.com/levitsky/pyteomics/pull/216)
- Enforce the grammar's maximum of two N- or C-terminal modifications.
- Correctly retain a valid second C-terminal modification.
- Reject text after a completed C-terminal modification or charge-adduct list.
- Reject empty peptidoforms and inputs ending in an incomplete grammar state.

For example, `[UNIMOD:5]-[UNIMOD:385]-PEPTIDE` previously parsed by silently losing the first N-terminal modification. It now raises `ProFormaError`; the valid spelling is `[UNIMOD:5][UNIMOD:385]-PEPTIDE`.

## Tests

- Added valid and invalid ProForma grammar-conformance matrices in `tests/test_proforma.py`.
- Ran `python -m unittest tests.test_proforma -q`.

### DIFF
--- a/pyteomics/proforma.py
+++ b/pyteomics/proforma.py
@@ -2879,7 +2879,8 @@ class Parser:
                 elif self.state == TAG_BEFORE:
                     self.state = POST_TAG_BEFORE
                 elif self.state == TAG_AFTER:
-                    self.c_term = self.current_tag()
+                    tags = self.current_tag()
+                    self.c_term.extend(tags)
                     self.state = POST_TAG_AFTER
                 elif self.state == GLOBAL:
                     self.state = POST_GLOBAL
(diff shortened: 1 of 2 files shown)


mobiusklein:
Thank you for the report and this initial work. This parser was written prior to 2.1, which did not have a formal grammar, and it has been kept that way to preserve previously accepted inputs. This particular scenario isn't quite what I had in mind though. 

The grammar you've based your work on also doesn't appear to be correct either, the specification itself doesn't place a limit on the number of modifications anywhere, and the validator at https://proforma.psidev.info/ accepts 3+ terminal modifications. I've opened an issue https://github.com/HUPO-PSI/ProForma/issues/33 at the main repository. The section in the specification that's been or being reviewed is https://docs.google.com/document/d/1HaSOMNmdFCxy0G6lq1Yci03z-4esmQoe_hSyoqTeRnc/edit?tab=t.0#heading=h.4f1mdlm

What I think we should do instead is either:
1) Emit a warning and skip the second terminal modification list intentionally. Preserving backwards compatibility, but update the internal state machine to do something similar to what you've proposed, but still less rigid. I think using the state of the various tag buffers as an ersatz memory for the state machine is probably the least intrusive way to go.
2) Emit a warning and merge the multiple modification lists for now, so that users know this abuse of notation is going away
3) as (1) but accept this particular notation is not the kind of input the Robustness Principle was intended to cover, so make this particular thing an error. 

The validation logic in `Parser._finish_component` does look like a good idea regardless.

BioGeek:
With https://github.com/HUPO-PSI/ProForma/commit/ccda3cbb1acddba7e7f36cb03abf5e9f5f880bb2 merged into [HUPO-PSI/ProForma](https://github.com/HUPO-PSI/ProForma)'s `master` branch, I have now updated this PR accordingly:

  - Removed the two-tag limits for N- and C-terminal modification lists.
  - Added 3-tag terminal cases to the conformance matrix.
  - Added the two positive examples introduced by upstream issue [#33](https://github.com/HUPO-PSI/ProForma/issues/33)’s fix.
  - Kept repeated N-terminal groups (`[+1]-[+2]-PEPTIDE`) as invalid.

mobiusklein:
Thank you. I'm happy with the state this is in, and it has also prompted me to revisit my existing philosophy of robustness vs. "bug-for-bug" compatibility to at least temper my enthusiasm for it.

